PR #4, opened 2026-03-05
[Spec] Initial commit of the first clauses

llvm-beanz:
This commit adds initial LaTeX text for the introduction, scope and normative references clauses. This adds the basic glossary as well as formatting macros and a CMake-based build script for generating HTML and PDF builds of the specification.


llvm-beanz:
I've attached a PDF and zip of the HTML site for reviewer convenience. The PDF generation seeks to conform to Ecma's house style, although not all macros are fully exercised in its current state. The proper Ecma title pages are still missing but will come in a subsequent change.

The site contents are pretty bare, but the image below is the HTML rendering applied to the current state of the specification text in Microsoft's [hlsl-specs](https://github.com/microsoft/hlsl-specs/) repository:
Screenshot 2026-03-05 at 12 53 03 PM


[hlsl.pdf](https://github.com/user-attachments/files/25777015/hlsl.pdf)
[Site.zip](https://github.com/user-attachments/files/25777016/Site.zip)

.gitignore
```diff
@@ -0,0 +1,2 @@
+_site/
+spec/svg-inkscape/
```
jessebarker, line 2: Do we want to add rules here for the CMake generated files as well as the actual build artifacts (`spec/html/` and `spec/hlsl.*`)?

jessebarker (approved):
LGTM.  Not sure if we care about documenting build environment.  For example,  for my macOS environment, I had to:
- brew install --cask mactex
- brew install inkscape
- brew install pandoc

.gitignore
llvm-beanz: We could. I generally build the artifacts into a separate "build" directory. The reason I have the Inkscape stuff here is because I was unable to figure out how to get it to write to an intermediate directory instead of right into the source folder 😞 .

llvm-beanz:
> LGTM. Not sure if we care about documenting build environment. For example, for my macOS environment, I had to:
> 
> * brew install --cask mactex
> * brew install inkscape
> * brew install pandoc

This is a good call. I'll add a readme listing the dependencies for building the spec in different formats.

farzonl:
> > LGTM. Not sure if we care about documenting build environment. For example, for my macOS environment, I had to:
> > 
> > * brew install --cask mactex
> > * brew install inkscape
> > * brew install pandoc
> 
> This is a good call. I'll add a readme listing the dependencies for building the spec in different formats.

Not to overcomplicate things but I've been wondering if web editor support might solve this problem of support across OSes.  Overleaf has github integration: https://docs.overleaf.com/integrations-and-add-ons/git-integration-and-github-synchronization/github-synchronization

It also has multiple collaborator support. Problem is it isn't free. 

The other alternative I was considering was codespaces (https://github.com/features/codespaces). We could setup the dev enviornment for building the spec in the container and vscode web client can already display pdfs so we could easily build and see the pdf changes

farzonl (approved):

farzonl:
should we add a `.gitattributes` file and preemptively solve the crlf issues before they start showing up?

llvm-beanz:
> should we add a `.gitattributes` file and preemptively solve the crlf issues before they start showing up?

Just noticed I already did that:
https://github.com/hlsl-tc57/tc57/blob/main/.gitattributes

.gitignore
jessebarker: It's fine either way.  I think if we have info in the readme or move the build to a container environment later, this won't matter.

llvm-beanz: I'm inept at docker, but it does seem ideal to have a container environment for this. In a follow-up PR after this one merges I'm going to add some GitHub actions jobs like we have over on the hlsl-specs repo (https://microsoft.github.io/hlsl-specs/). We will have a post-merge build job to build the spec and webpage and publish them to GitHub pages, and a PR job to build before merge so we can have changes tested and artifacts available for review.

llvm-beanz:
This was reviewed and approved for merging in the 2026/03/31 meeting.

PR #8, opened 2026-03-19
Add Terms and Definitions Clause

llvm-beanz:
This PR adds the Terms and Definitions Clause with a bunch of useful definitions that will be used throughout the specification. This is not intended to be a complete set of terms and definitions, but should be a reasonable starting point.


llvm-beanz:
Note to reviewers: This PR is stacked on top of PR #4 because that PR has all the infrastructure and basic structure of the specification text. Once that PR merges I will merge it into this branch so that the changes here are easier to review. Until then, the main text to review here is the added `terms.tex` file.

spec/terms.tex
```diff
@@ -0,0 +1,223 @@
...
+\p \acrshort{hlsl} is a \acrfull{spmd} programming language. The following terms
+are used to describe the execution model of \acrshort{spmd} programs and the
+semantics of language constructs that are specific to \acrshort{spmd}
+programming.
+
+\Sub{\gls{lane}}{Terms.Lane}
+
+\p A \gls{lane} represents a single computed element in an \acrshort{spmd}
+program. In a traditional programming model it would be analogous to a thread of
+execution, however it differs in one key way. In multi-threaded programming
```
s-perron:
```suggestion
execution, however it differs in one key way. In multi-threaded programming,
```

spec/terms.tex
```diff
@@ -0,0 +1,223 @@
...
+output results based on the initial launch conditions and program control flow.
+
+\p A \textit{helper} \gls{lane} is a lane which would not be executed by the
+initial launch conditions except that its computations are required for adjacent
+pixel operations in pixel fragment shaders. A \textit{helper} \gls{lane} will
+execute all computations but will not perform writes to buffers, and any outputs
+it produces are discarded. \textit{Helper} lanes may be required for
+\gls{lane}-cooperative operations to execute correctly.
+
+\p A \textit{inactive} \gls{lane} is a lane that is not executed by the initial
```
s-perron:
```suggestion
\p An \textit{inactive} \gls{lane} is a lane that is not executed by the initial
```

spec/terms.tex
```diff
@@ -0,0 +1,223 @@
...
+    \path[->] (helper) edge node {branch} (off2);
+    \path[->] (off2) edge node {} (helper);
+  \end{tikzpicture}
+\end{figure}
+\end{center}
+
+\Sub{Uniformity}{Terms.Uniformity}
+
+\p Uniformity is a property of either a value or a control flow construct.
+Uiniformity can be discussed in terms of any grouping of \gls{lane}s (e.g.
```
s-perron:
```suggestion
Uniformity can be discussed in terms of any grouping of \gls{lane}s (e.g.
```

spec/terms.tex
```diff
@@ -0,0 +1,223 @@
...
+\gls{quad}, \gls{threadgroup}, \gls{wave}, \gls{dispatch}), if a grouping of
+\gls{lane}s is not explicitly mentioned, it is assumed to be a \gls{wave}.
+
+\p When referring to a value, uniformity means that the value is the same across
+all \gls{lane}s in the grouping. When referring to a control flow construct,
```
s-perron, line 199: Can a value computed in non-uniform control flow be considered uniform?

```
if ( i == 10) {
  // Not every thread will have the same value for `Samplers[i]`, but every active thread will.
  tex.Sample(Samplers[i], uv);
}
```

I'm trying to figure out how this defintion should map to SPIR-V's [dynamically uniform](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#DynamicallyUniform).

This might be important for how we define function parameters like the texture sample function.

llvm-beanz: Yea, I was trying to get that a few lines down where it says:

> Uniformity can be a static property that is determined at compile time, or a dynamic property that is determined at runtime.

If you have suggestions for how to more clearly express this please let me know.

s-perron: I changed my mind. We can leave this as it for now, but we may have to make the definition more precise or add new terms when we see where the term is used.

spec/hlsl.tex
```diff
@@ -58,6 +58,8 @@
 \clearpage
 
 % Switch to arabic numbering starting at 1 for main content
```
s-perron: You might want to move this comment to be with the `pagenumbering` change.

s-perron (approved):

PR #11, opened 2026-03-29
Add Conformance clause

llvm-beanz:
This adds a basic conformance clause. This will likely evolve as we flesh out the specification if we add optional features, forbidden features, or a conformance test suite.

This PR is currently based on top of #4. Once that PR merges this will be rebased to minimize the diffs.


spec/conformance.tex
```diff
@@ -0,0 +1,16 @@
...
+\p A conforming implementation of \acrshort{hlsl} must provide and support all
+the types, values, objects, attributes, functions, and program syntax and
+semantics described in this specification.
+
+\p A conforming implementation of \acrshort{hlsl} may provide additional types,
+values, objects, attributes, functions beyond those described in this
+specification, provided that they either do not modify the base namespace and
+namespaces defined in this specification, or if they do modify those namespaces,
+the modifications are prefixed with the reserved \textit{double underscore}
+(\texttt{__}) prefix.
```
bogner: It's somewhat pedantic because I don't expect to define anything in the spec with reserved names, but I don't like the wording that "modifications are prefixed". I think the exception we want to carve out is specifically "additions that are prefixed". Also what about undescore-capital (`_[A-Z]`) reserved words? Are those allowed?

Here's an attempt:
```suggestion
\p A conforming implementation of \acrshort{hlsl} may provide additional types,
values, objects, attributes, and functions beyond those described in this
specification. Additions must not be part of the base namespace or any
namespaces defined by this specification unless they are implementation details
with names that use the reserved \textit{double underscore} (\texttt{__})
prefix.
```

spec/conformance.tex
```diff
@@ -0,0 +1,16 @@
...
+
+\p A conforming implementation of \acrshort{hlsl} must not redefine any behavior
+that is not defined as implementation-defined, undefined or unspecified in this
+specification.
```
bogner, line 17: This reads to me like a bit of a confusing double negative. Here's an attempt to clarify it, though I'm not entirely convinced it's very clear either.

```suggestion
\p A conforming implementation of \acrshort{hlsl} must not redefine any
behavior from this specification. Note that bahevior that is
implementation-defined, undefined, or unspecified is not considered to be
defined by this specification.
```

spec/conformance.tex
llvm-beanz: One of the things my original language allowed that your revision doesn't is (for example) if we have an enumeration defined in the specification, the original language would allow adding additional cases to the enumeration that are implementation-provided.

What I'm trying to allow is cases where API-specific features that are not implementation details may need to be added into existing types or namespaces in order to expose new functionality.

spec/conformance.tex
llvm-beanz: I'm not sure your rewrite is clearer to understand. It is more natural language, but I worry that it reads more like a sentence that can't stand by itself and a second sentence to weaken the first.

FWIW, I took that sentence almost word-for-word from the ECMAScript conformance clause:

> A conforming implementation of ECMAScript must not redefine any facilities that are not [implementation-defined](https://tc39.es/ecma262/#implementation-defined), [implementation-approximated](https://tc39.es/ecma262/#implementation-approximated), or [host-defined](https://tc39.es/ecma262/#host-defined).

[Source](https://tc39.es/ecma262/#sec-conformance)

spec/hlsl.tex
```diff
@@ -62,8 +62,9 @@
 % Switch to arabic numbering starting at 1 for main content
 \pagenumbering{arabic}
 \input{scope}
-\input{terms}
+\input{conformance}
 \input{references}
+\input{terms}
```
llvm-beanz, line 67: This is a small editorial fix to the ordering of clauses. The first three clauses should be (in order) scope, conformance, and normative reverences as per the Ecma house style rules. I spotted that we had terms and definitions out of order when merging in main to resolve the conflict.

spec/conformance.tex
bogner: I think that allowing implementations to add values to existing enums is a dangerous idea - such could end up colliding with future standard values in incompatible ways. With my wording this would be disallowed, and an implementation that wanted an extended enum would have to duplicate the standard enum (plus additional values) in their own namespace.

spec/conformance.tex
bogner: I still find the original hard to follow, but that might just be me. The wording as you have it is fine.

farzonl (approved):

spec/conformance.tex
farzonl: This sounds like we resolved things and should be able to merge by or before tomorrows meeting?

spec/conformance.tex
llvm-beanz: It wouldn't let me apply your suggestion directly because of other changes to the file, but I've applied your recommended language. I think we can safely start more restrictive and relax if needed.

bogner (approved):

PR #15, opened 2026-04-03
[New] Conforming literals proposal

llvm-beanz:
This introduces the conforming literals proposal that aligns HLSL's handling of literal values more closely with C and C++.

This feature was planned and implemented by the HLSL team as part of the HLSL 202x feature set, and is fully implemented and tested in both DXC and Clang.


farzonl:
@pow2clk ping for review

devshgraphicsprogramming:
This is a mess for us in  a shared C++ and HLSL codebase without the support for C++23 `f16`, `f32`, `f64` floating point literal suffices.

We started needing `NBL_FLOAT64_LITERAL` macro after the corresponding change in DXC to make sure we get correct width literals.

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+#### Methodology
+
+A modified version of DXC implementing the new behavior enabled under all
+language modes is available
+[here](https://github.com/llvm-beanz/DirectXShaderCompiler/tree/cbieneman/conforming-literals).
+This modified version of DXC was used as the compiler for testing to identify
+the impact of this change on existing shaders used in production software.
+
+Tests were performed by both Microsoft and Google using private test suites. In
+Google's testing, *all tests passed all tests with no issues*.
```
s-perron: Our testing is very small. I don't think it should be used a evidence that this changes does not cause issues.

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+`asfloat16` is a barrier for users. Further discussion is needed to determine
+how best to address this problem.
+
+### Mitigating Migration Pain
+
+There are no observed instances of the change in floating point behavior causing
+disruption. If such cases exist the existing `-Wconversion` diagnostics will
+sufficiently notify users of implicit conversions that result in precision loss.
+
+The change in integer behavior inc more important. The behavior of bit shifts in
```
s-perron:
```suggestion
The change in integer behavior is more important. The behavior of bit shifts in
```

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+HLSL has historically changed independent of language version, and ambiguity of
+bit shifts on literal operands does produce a compiler diagnostic (see:
+[dxc/#300](https://github.com/microsoft/DirectXShaderCompiler/pull/3400)).
+
+This proposal will introduce a new diagnostic for 32-bit literal integers values
+that will become `unsigned` with this proposal. This diagnostic will be under
+the `-Wfuture-compatability` diagnostic group.
```
s-perron:
```suggestion
the `-Wfuture-compatibility` diagnostic group.
```

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+[#175](https://github.com/microsoft/hlsl-specs/pull/175).
+
+The full proposed specification for integer literals in HLSL is in
+[#208](https://github.com/microsoft/hlsl-specs/pull/208).
+
+Documentation for related conversion rank behavior is in the **[Conv.rank]**
+section of the language spec (minimum precision types added in
+[#206](https://github.com/microsoft/hlsl-specs/pull/206)).
+
+### Notes on minimum precision types
```
s-perron, line 267: Will there be another proposal defining the minimum precision types? I ask because you want this PR to incorporate https://github.com/microsoft/hlsl-specs/pull/206, which modified existing language around `min16*` types. 

In particular, it says 

> This means when stored to memory
objects of type \texttt{min16float} are stored as \textbf{binary32} as defined
in \gls{IEEE754}.

That is not what is implemented in DXC when 16-bit types are enabled.

Should we discuss this now or later?

s-perron (commented):
I'm in favour of this proposal. Just a couple minor changes.

proposals/0002-conforming-literals.md
llvm-beanz: The minimum precision types may not be used when 16-bit types are enabled, so they're exclusive to native 16-bit types.

I don't so much expect we'll have a proposal for defining minimum precision types since they're already part of the implementation I expect that we'll just directly draft spec language defining how they behave today rather than a proposal to change their behavior.

llvm-beanz (commented):
As a meta-comment to myself, I think the compiler warnings and associated flags shouldn't be part of the specification language as they're really implementation details for _an_ implementation not requirements for conformance.

I'll revise the document to clarify that.

proposals/0002-conforming-literals.md
s-perron: Okay, now I see the warning in DXC that the min16float will be "promoted to half".

farzonl:
Are `proposals/NNNN-assets/DiffExample1.png` and `proposals/NNNN-assets/DiffExample2.png` supposed to be black images?

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+### Experimentation
+
+In writing this proposal a key thesis was proposed that the impact of this
+change on developers would be minimal because DXC's behavior for DXIL is
+undocumented, complex, and different from other shader compilers (including
+DXC's SPIR-V generator). The last point is important and relevant because a
+large percentage of shader code is compiled with multiple different shader
+compilers which implement different language semantics. Because of this, it is
+believed that many existing shaders are engineered to be resilient to subtle
+behavioral differences like the ones this change would cause.
```
farzonl, line 112: It sounds like this is unlikely to be an issue if the resiliency is true. But also many of these tools are a DXIL to X translation. This proposal is in the frontend seems like an odd call out.

 But if it isn't why does the experimentation responsibility fall on our shoulders? If things break, I would expect maintainers of those other projects to update their tools.

farzonl:
> Are `proposals/NNNN-assets/DiffExample1.png` and `proposals/NNNN-assets/DiffExample2.png` supposed to be black images?

Nevermind, I can see one purple pixel in one of the images.

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+  [microsoft/DirectXShaderCompiler#6410](https://github.com/microsoft/DirectXShaderCompiler/issues/6410),
+  [shader-slang/slang#1185](https://github.com/shader-slang/slang/issues/1185)
+
+## Introduction
+
+In C-based languages literals are tokens which the compiler interprets in the
+most basic translation to preserve the exact meaning expressed in the source to
+the final program.
+
+HLSL's handling of literals is complex, undocumented, and inconsistent.
```
pow2clk: I try to write these sorts of documents such that the reflect the state of affairs after they are accepted. Hopefully this is no longer true once this is accepted and leaving it here will have it during the majority of the document's lifespan. 

Saying prior to this proposal that was the case seems okay for the motivation section, but I think what's there already conveys that with more detail

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+### Benefits of this solution
+
+The most clear and obvious benefits of this solution are its simplicity. The
+implementation of this behavior is just a few lines of code. The full
+specification is simple and concise.
+
+This solution also works with modern C++ features that have come into HLSL like
+templates, and other features like `auto` which we would like to add. It
+addresses issues like the bugs with the [ternary
+operator](microsoft/DirectXShaderCompiler#6147), where a comprehensive solution
```
pow2clk: This link doesn't work. Needs the full URL
```suggestion
operator](microsoft/DirectXShaderCompiler#6147), where a comprehensive solution
operator]((https://github.com/microsoft/DirectXShaderCompiler#6147), where a comprehensive solution
```

proposals/0002-conforming-literals.md
pow2clk: Whose shoulders are "ours" here? 

Overall, I think this might be a bit difficult to read. I'd like to get some confirmation that I understand what's being said.

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Microsoft Test Suite
+
+The test cases are PIX captures running on Xbox Series X. Most of the captures
+are rendered frames of in-development or shipped games, or sample content
+produced from external developers working on games or engine technology. The
+remaining captures are test programs developed internally at Microsoft.
+
+For the test runs the image difference tolerance was zero, meaning that any
+frame which had even a single bit difference in output was marked as a failure.
```
pow2clk, line 134: Could this be further tested in the offload test suite on various hardware?

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+* Extremely minor - 8 test cases
+* Minor - 3 test cases
+* Subtle visible differences - 2 test cases
+* Catastrophic visible differences - 6 test cases
+
+##### Compile Failures
+
+All 10 shaders that failed to compile hit a variation of the issue reported as
+[DXC/#6315](https://github.com/microsoft/DirectXShaderCompiler/issues/6315). The
```
pow2clk, line 150: a cast does solve this, but perhaps f16 suffixes would be helpful as well as was suggested in comment https://github.com/hlsl-tc57/tc57/pull/15#issuecomment-4250137738

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+overall image still rendered mostly correctly.
+
+##### Catastrophic Visible Differences
+
+We have 6 test cases, all from the same software title which had severe
+obviously incorrect rendering which would make the title unplayable. One
+important note about this title is that it is an Xbox & Windows exclusive title.
+Since the title is platform exclusive its shaders may not need to be resilient
+to different language semantics implemented by different shader compilers.
+
```
pow2clk, line 190: I'm not sure how useful this rundown of results is given how limited the pool is. I think the acknowledgement that catastrophic differences are possible with the explanation as to why is sufficient to convey what the possibilities are. It all depends on how the shaders in question are written and focusing on what is most likely to cause differences is more useful than listing the outcomes of a handful of results available.

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+* If `L` or `UL` suffix is present, it will parse a literal larger than
+  UINT32_MAX , though it will be truncated to 32-bits when used, since there is
+  no 64-bit integer support in DXBC.
+* FXC has unexpected behavior in corner cases for constant evaluation, such as
+  overflow from multiply resulting in `INT32_MIN` `((int)0x80000000)`
+
+#### Conclusions Drawn
+
+The testing here holds with the core thesis that the number of impacted software
+titles will be minimal and that most shaders are resilient to the changes in
```
pow2clk, line 238: I would not draw that conclusion from available evidence. I would instead say that it depends on how the shaders are written and adopters should investigate that before switching to 202x

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+precision. The presence of catastrophic rendering defects does indicate that
+there may be severe cases where an individual developer may need to put
+significant effort in to resolve issues that arise. Further since the only title
+to experience catastrophic rendering defects is an Xbox and Windows exclusive
```
pow2clk, line 242: SPIR-V generation observes the same catastrophic switch from arithmetic to logical shift using a slightly modified shader to get around its linkage limitations: https://godbolt.org/z/5GGo8Eadf So shaders that rely on that behavior and target multiply platforms are possible.

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+integer, whereas in C a non-base 10 literal integer may be unsigned if the most
+significant bit is 1.
+
+This difference in rules results in a subtle behavioral difference for bit
+shifts. Given the following code:
+
+```c++
+export float Fn(int inInt, inFloat) {
+  int Val = (inInt & 0xffff0000) >> 16);
+  return Val* inFloat
```
pow2clk: Some typos
```suggestion
export float Fn(int inInt, float inFloat) {
  int Val = (inInt & 0xffff0000) >> 16;
  return Val* inFloat;
```

proposals/0002-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+title, the reasoning behind the thesis that most shaders are resilient due to
+required compatibility with other shader compilers seems a likely factor.
+
+The testing also revealed that the lack of complete overload sets for
+`asfloat16` is a barrier for users. Further discussion is needed to determine
+how best to address this problem.
```
pow2clk, line 248: This text was originally written two years ago. Has that discussion taken place, might this be the time for it?

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+sufficiently notify users of implicit conversions that result in precision loss.
+
+The change in integer behavior is more important. The behavior of bit shifts in
+HLSL has historically changed independent of language version, and ambiguity of
+bit shifts on literal operands does produce a compiler diagnostic (see:
+[dxc/#300](https://github.com/microsoft/DirectXShaderCompiler/pull/3400)).
+
+This proposal will introduce a new diagnostic for 32-bit literal integers values
+that will become `unsigned` with this proposal. This diagnostic will be under
+the `-Wfuture-compatibility` diagnostic group.
```
pow2clk: It seems to have been named `-Wfuture-compat` and it doesn't fire in the catastrophic case above https://godbolt.org/z/bhehfG99Y The only uses are for initialization warnings.

proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+This proposal will also add a new diagnostic for calls to `asfloat16` with
+floating point arguments to notify users that they should use a cast instead.
```
pow2clk: This is also still to be implemented: https://godbolt.org/z/eGK3YGKdn

pow2clk (commented):
In spite of all my critiques, I do support this proposal. I think its current form inaccurately plays down the risks in a way that will deter developers from allocating sufficient time to perform the transition to the new language version. 

I'm aware that there is talk of removing the mention of warnings and such. I think they were originally included to further downplay the risks and give assurance that there would be tools to smooth the way. Given that the most dangerous changes still have no warnings in DXC, I think that needs to be addressed and acknowledged rather than some pat statement that DXC will include helpful warnings where necessary unless there is a plan to actually add those warnings

proposals/0002-conforming-literals.md
llvm-beanz: The behavior of literals is completely handled by the compiler, it (by definition) can't vary across hardware except in cases where hardware is allowed to behave differently with the results of a literal (things like floating point precision).

We already have many tests in the offload-test-suite testing literals using the behavior described in this proposal, but that doesn't give us a signal about the impact of this on existing HLSL codebases since those tests were all crafted after this proposal was implemented in Clang.

proposals/0002-conforming-literals.md
llvm-beanz: Tracking this: https://github.com/hlsl-tc57/tc57/issues/31

proposals/0002-conforming-literals.md
llvm-beanz: Even though our test suite was not extensive, we have in the past found it to be a reasonable representation. For that reason I found it valuable to both describe the size of the suite and the number of failures of different kinds to provide complete context. This analysis was also performed at the request of other members of the HLSL team who are now delegates to the TC, so I suspect keeping it here for posterity is valuable.

proposals/0002-conforming-literals.md
llvm-beanz: The testing did not invalidate the core thesis, therefore it held with the thesis. It does not _prove_ the thesis, but it also does not disprove it.

proposals/0002-conforming-literals.md
llvm-beanz: Tracking as: https://github.com/hlsl-tc57/tc57/issues/32

proposals/NNNN-conforming-literals.md
llvm-beanz: `-Wfuture-compat` is an unrelated warning for legacy C++ behaviors.

As per the discussion during the meeting, and my comment on the PR (https://github.com/hlsl-tc57/tc57/pull/15#pullrequestreview-4165530632), we should remove discussion of compiler diagnostics since they are outside the scope of the specification.

llvm-beanz:
> Are `proposals/NNNN-assets/DiffExample1.png` and `proposals/NNNN-assets/DiffExample2.png` supposed to be black images?

They're actually not fully black... the only pixels with color are the ones that differed between the frames rendered with this change and without it. They're largely imperceptible, which is the illustrating point.

proposals/NNNN-conforming-literals.md
llvm-beanz: Also as a side note, the hlsl-specs proposal this was copied from contains conflicting language whether this warning is part of the `-Wfuture-compatabiltiy` diagnostic group or the `-Whlsl-legacy-literal` group.

The later is implemented (see: https://godbolt.org/z/no5K98vYr).

llvm-beanz:
In the meeting this week we had consent to merge this after fixing links, removing the warnings-related text, and filing issues for larger outstanding issues. I've assigned this proposal 0002 in preparation for merging.

@pow2clk can you confirm that you're okay with this merging?

llvm-beanz:
ping @pow2clk

llvm-beanz:
ping @pow2clk

farzonl (approved):

pow2clk:
Sorry I didn't respond to this. I need to change where my email notifications go. I confirm I'm okay with it going in.

PR #19, opened 2026-04-14
[New] Proposal for C++-aligned overload resolution

llvm-beanz:
This proposal suggests adoptiong C++-aligned overload resolution behavior which will be more intuitive for users, and less complex for compiler implementations.

### DIFF
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -1,5 +1,5 @@
 ---
-title: "NNNN - Simplified Overload Resolution"
+title: "0003 - Simplified Overload Resolution"
 draft: true
 params:
   authors:
--- a/proposals/NNNN-overload-resolution.md
+++ b/proposals/NNNN-overload-resolution.md
@@ -44,7 +44,9 @@ This proposal suggests introducing new implicit conversions and conversion ranks
 to classify conversions involving matrix and vector types as source and/or
 destination arguments. The table below is the full list of proposed implicit
 conversions and their ranks. The **bolded** are unique to HLSL and introduced by
-this proposal, the remaining elements are inherited from C++.
+this proposal, the remaining elements are inherited from C++. The table below is
+sorted by the ranks in increasing order such that the conversion of the _lowest_
+rank is preferred.
 
 | Conversion  | Rank |
 |-------------|------|
@@ -66,10 +68,10 @@ this proposal, the remaining elements are inherited from C++.
 | **Component-wise conversion** | Conversion |
 | **Scalar splat conversion** | **Conversion Extension** |
 | **Vector truncation (without conversion)** | Truncation |
-| **Vector truncation promotion** | **Promotion Truncation** |
-| **Vector truncation conversion** | **Conversion Truncation** |
 | **Matrix truncation (without conversion)** | Truncation |
+| **Vector truncation promotion** | **Promotion Truncation** |
 | **Matrix truncation promotion** | **Promotion Truncation** |
+| **Vector truncation conversion** | **Conversion Truncation** |
 | **Matrix truncation conversion** | **Conversion Truncation** |
 
 The new conversions allow combining element type conversions with dimension
--- a/proposals/NNNN-overload-resolution.md
+++ b/proposals/NNNN-overload-resolution.md
@@ -1,0 +1,112 @@
+---
+title: "NNNN - Simplified Overload Resolution"
+draft: true
+params:
+  authors:
+    - llvm-beanz: Chris Bieneman
+  sponsors:
+    - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+This proposal suggests adopting an overload resolution algorithm that aligns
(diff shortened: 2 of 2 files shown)


farzonl:
pow2clk is intrested in reviewing this

proposals/0003-overload-resolution.md
```diff
@@ -0,0 +1,112 @@
...
+| Floating-integral conversion | Conversion |
+| Boolean conversion | Conversion |
+| **Component-wise conversion** | Conversion |
+| **Scalar splat conversion** | **Conversion Extension** |
+| **Vector truncation (without conversion)** | Truncation |
+| **Vector truncation promotion** | **Promotion Truncation** |
+| **Vector truncation conversion** | **Conversion Truncation** |
+| **Matrix truncation (without conversion)** | Truncation |
+| **Matrix truncation promotion** | **Promotion Truncation** |
+| **Matrix truncation conversion** | **Conversion Truncation** |
```
devshgraphicsprogramming, line 75: IMHO truncation and promotion should be behind a flag and warn. 

Slang doesn't let you do as many conversions and promotions and thats a good thing.

llvm-beanz: Whether or not they warn is outside the scope of the language specification (although in Clang they do quite nicely).

I would also prefer that we not support these conversions implicitly, and would like them to only be allowed through explicit conversion operators, however that would almost certainly break a lot of existing HLSL code.

This proposal seeks to minimize how much we break existing code by allowing all the conversions that are currently allowed. We're restricting the changes in this proposal to just how the overload set is determined, where overload sets are determined, and how the best candidate function is chosen.

proposals/0003-overload-resolution.md
```diff
@@ -0,0 +1,112 @@
...
+| Integral promotion | Promotion |
+| Floating point promotion | Promotion |
+| **Component-wise promotion** | Promotion |
+| **Scalar splat promotion** | **Promotion Extension** |
+| Integral conversion | Conversion |
+| Floating point conversion | Conversion |
+| Floating-integral conversion | Conversion |
+| Boolean conversion | Conversion |
+| **Component-wise conversion** | Conversion |
+| **Scalar splat conversion** | **Conversion Extension** |
```
pow2clk, line 69: How does this differ from the splat to vector and splat to matrix listed above? Perhaps it involves element type conversion as well? Maybe those should be listed with promotion and conversion variants as with the truncations below?

proposals/0003-overload-resolution.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Implicit conversions
+
+This proposal suggests introducing new implicit conversions and conversion ranks
+to classify conversions involving matrix and vector types as source and/or
+destination arguments. The table below is the full list of proposed implicit
+conversions and their ranks. The **bolded** are unique to HLSL and introduced by
+this proposal, the remaining elements are inherited from C++.
+
+| Conversion  | Rank |
```
pow2clk, line 51: I'm aware of these ranks being listed in descending order of preference in https://github.com/microsoft/hlsl-specs/blob/c4b9b88b0ca4eedd3ee4a2ba3e06b4e3f926270a/specs/language/overloading.tex#L339 What is the mechanism for merging this with that? Can we reference or even repeat that here for clarity?

pow2clk (commented):
Apart from the two requests for clarification, I don't see much to object to here, but it feels a bit thin. Overload resolution algorithms are referred to, but don't seem fully fleshed out. Is the intent to use this as a starting point?

proposals/NNNN-overload-resolution.md
```diff
@@ -0,0 +1,112 @@
...
+
+| Conversion  | Rank |
+|-------------|------|
+| No conversion (Identity) | Exact Match |
+| Lvalue-to-rvalue | Exact Match |
+| Array-to-pointer | Exact Match |
+| Qualification | Exact Match |
+| **Single Element to Scalar** | Exact Match |
+| **Vector Scalar splat conversion** | **Conversion Extension** |
+| **Matrix Scalar splat conversion** | **Conversion Extension** |
```
llvm-beanz:
```suggestion
| **Vector Scalar splat conversion** | **Extension** |
| **Matrix Scalar splat conversion** | **Extension** |
```

pow2clk (approved):
My concerns are resolved

proposals/0003-overload-resolution.md
llvm-beanz: This will all be populated in the detailed design as the proposal moves through the lifecycle. For now I'll add a note about the ranks being listed in increasing order in the table.

proposals/0003-overload-resolution.md
llvm-beanz: I addressed this in the meeting, the table was incorrect the earlier listed conversions are **Extension** not **Conversion Extension**.

llvm-beanz:
In the 2026-04-28 TC 57 meeting we reached agreement to merge this as **Under Consideration**. This proposal is assigned number 0003, and I will update the PR accordingly.

farzonl (approved):

PR #23, opened 2026-04-27
Add Basic Concepts clause

llvm-beanz:
This adds the Basic Concepts clause and the first few sub-clauses to provide some basic background around declarations and definitions and a note inheriting C++'s One-Definition Rule.


spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+  \p HLSL inherits a significant portion of its language semantics from C and
+  C++. Some of this is a result of intentional adoption of syntax early in the
+  development of the language and some a side-effect of the Clang-based
+  implementation of DXC.
+
+  \p This chapter includes a lot of definitions that are inherited from C and
+  C++. Some are identical to C or C++, others are slightly different. HLSL is
+  neither a subset nor a superset of C or C++, and cannot be simply described in
+  terms of C or C++. This specification includes all necessary definitions for
+  clarity.
```
pow2clk: I think I understand this last statement to say that it includes both the C/C++ derived statements with any necessary slight differences as well as those that are original to HLSL?

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+\Sec{Preamble}{Basic.preamble}
+
+\p An \textit{entity} is a value, object, function, enumerator, type, class
+member, bit-field, template, template specialization, namespace, or pack.
+
+\p A \textit{name} is a use of an \textit{identifier} (\ref{Expr.Primary.ID}),
+\textit{operator-function-id} (\ref{Overload.operator}),
+\textit{conversion-function-id} (\ref{Classes.Conversions}),
+or \textit{template-id} (\ref{Template}) that denotes any entity or
+\textit{label} (\ref{Stmt.Label}).
```
pow2clk, line 22: Some of these refs don't exist yet. I don't mind how we want to approach this, but since this isn't exactly the same as the other version of this file, I know it saw some likely temporary redactions and I don't know if we want to leave these in for later convenience or not.

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+
+\p Every name that denotes an entity is introduced by a \textit{declaration}.
+Every name that denotes a label is introduced by a \textit{labeled statement}
+(\ref{Stmt.Label}).
```
pow2clk, line 26: I liked the clarifying footnote that specified labels are only for switch statements in the original, but I won't insist on it.

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+
+\p A \textit{variable} is introduced by the declaration of a reference other
+than a non-static data member of an object. The variable's name denotes the
+reference or object.
+
+\p Whenever a name is encountered it is necessary to determine if the name
+denotes an entity that is a type or template. The process for determining if a
+name refers to a type or template is called \textit{name lookup}.
```
pow2clk: The wording here makes it sound like the lookup is to determine which of these possibilities are true:
a. It's a type
b. It's a template

I think instead it's meant to determine between:
a. It's a type or template
b. It's something else

But I guess this is how it's written in isoCPP?

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+\item use of attributes (\ref{Decl}), and
+\item controlling template instantiation (\ref{Template.Inst}).
+\end{itemize}
+
+\p A declaration is a \textit{definition} unless:
+\begin{itemize}
+\item it declares a function without specifying the function's body
+(\ref{Decl.Function}),
+\item it is a parameter declaration in a function declaration that does not
+specify the function's body (\ref{Decl.Function}),
```
pow2clk, line 61: I don't know what parameters specifying the fucntion's body means

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+\item it is a global or namespace member declaration without the \texttt{static}
+specifier,
+\item it declares a static data member in a class definition,
+\item it is a class name declaration,
+\item it is a template parameter,
+\item it is a \texttt{typedef} declaration (\ref{Decl}),
+\item it is an \textit{alias-declaration} (\ref{Decl}),
```
pow2clk, line 68: are there alias declarations that are not "using" declarations?

pow2clk (commented):
Looks good. Mostly just some clarifications that likely won't result in any changes.

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+}
+tbuffer TB {                 // does not declare TB
+  int w;                     // declares w
+}
+\end{HLSL}
+
+\Sec{One-Definition Rule}{Basic.ODR}
+
+\p The \gls{isoCPP} \textit{One-definition rule} is adopted as defined in
+\gls{isoCPP} \textbf{[basic.def.odr]}.
```
pow2clk, line 118: I can perhaps guess why some of the basics content was left out for now, but I think this section is needed: 
```suggestion
\gls{isoCPP} \textbf{[basic.def.odr]}.

-\Sec{Lvalues and rvalues}{Basic.lval}
-
-\p Expressions are classified by the type(s) of values they produce. The valid
-types of values produced by expressions are:
-
-\begin{enumerate}
-  \item An \textit{lvalue} represents a function or object.
-  \item An \textit{rvalue} represents a temporary object.
-  \item An \textit{xvalue} (expiring value) represents an object near the end
-  of its lifetime.
-  \item A \textit{cxvalue} (casted expiring value) is an \textit{xvalue}
-  which, on expiration, assigns its value to a bound \textit{lvalue}.
-  \item A \textit{glvalue} is an \textit{lvalue}, \textit{xvalue}, or
-  \textit{cxvalue}.
-  \item A \textit{prvalue} is an \textit{rvalue} that is not an \textit{xvalue}.
-\end{enumerate}
```

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+member, bit-field, template, template specialization, namespace, or pack.
+
+\p A \textit{name} is a use of an \textit{identifier} (\ref{Expr.Primary.ID}),
+\textit{operator-function-id} (\ref{Overload.operator}),
+\textit{conversion-function-id} (\ref{Classes.Conversions}),
+or \textit{template-id} (\ref{Template}) that denotes any entity or
+\textit{label} (\ref{Stmt.Label}).
+
+\p Every name that denotes an entity is introduced by a \textit{declaration}.
+Every name that denotes a label is introduced by a \textit{labeled statement}
```
pow2clk, line 25: Here in case you want it.
```suggestion
Every name that denotes a label is introduced by a \textit{labeled statement}

\begin{note}
HLSL does not have \texttt{goto}, and labeled-statements are only valid
within \texttt{switch} statements.
 \end{note}

```

pow2clk (commented):
Just some suggestions in case they are wanted

spec/basic.tex
```diff
@@ -0,0 +1,125 @@
...
+\item it is a \texttt{typedef} declaration (\ref{Decl}),
+\item it is an \textit{alias-declaration} (\ref{Decl}),
+\item it is a \textit{using-declaration} (\ref{Decl}),
+\item it is a \textit{static\_assert-declaration} (\ref{Decl}),
+\item it is an \textit{empty-declaration} (\ref{Decl}),
+\item or a \textit{using-directive} (\ref{Decl}).
+\end{itemize}
+
+\begin{note}
+  Global variable declarations are implicitly constant and external in HLSL.
```
jessebarker: This doesn't include `static` declarations, though, does it?

jessebarker: (at least the external part)

spec/basic.tex
llvm-beanz: Ecma house style does not allow for footnotes. Some footnotes I've transcribed into notes, but others I've chosen to remove if they didn't seem meaningful enough to give the page space of a full note.

spec/basic.tex
llvm-beanz: The LaTeX build does not broken references. My intent was to leave all the references in so they get filled in as the spec text moves over, and do a pass once we've finished integrating language from Microsoft's repository to fix up any broken labels at that point.

spec/basic.tex
llvm-beanz: Reading one paragraph in isolation is likely to be confusing. Types and templates are entities, but there's other things that are entities too (full list is a couple paragraphs above).

Do you have any alternate suggestions for how to clarify this wording?

C++ says:

> Some names denote types or templates. In general, whenever a name is encountered it is necessary to determine whether that name denotes one of these entities before continuing to parse the program that contains it. The process that determines this is called [name lookup](https://timsong-cpp.github.io/cppwp/n3337/basic#def:name_lookup) ([[basic.lookup]](https://timsong-cpp.github.io/cppwp/n3337/basic.lookup)).

https://timsong-cpp.github.io/cppwp/n3337/basic#7

spec/basic.tex
llvm-beanz:
```
void MyFunction(int X);
```

`X` is a parameter declaration in a function declaration that does not specify the function's body.

spec/basic.tex
llvm-beanz: No, that is an alias-declaration. They were introduced in HLSL 2021.

spec/basic.tex
llvm-beanz: I intentionally left this out because I was trying to phase in the specification text.

spec/basic.tex
llvm-beanz: I personally don't find this comment necessary. If we call out every difference from C++ in the spec we're going to have _a lot_ of language calling out differences.

@s-perron & @Alexander-Johnston do you have thoughts here?

spec/basic.tex
llvm-beanz: In order to address the fact that other PRs do rely on this language, I've posted the whole text of the basic clause from the Microsoft draft into this PR.

spec/basic.tex
Alexander-Johnston: I agree it's not necessary. I would rather that in general we make explicit what we include from C++ rather than what we don't. Especially for `goto`, given that we won't have a section defining it I think it'll be clear enough to readers that HLSL doesn't support it.

llvm-beanz:
@pow2clk, can you please review and provide any feedback?

spec/basic.tex
pow2clk: This is the only one that got removed. My first thought was "why does HLSL need labels?" I did remember switch cases eventually, but this would have prevented that initial confusion. As I said before, I won't insist on it.

spec/basic.tex
pow2clk: I see. It's the function declaration that does not specify the body. I'm not sure if there is better wording here or if I was just in the wrong state of mind when I read it initially as I couldn't imagine how a parameter would specify a function body. I'll think about it a bit more

spec/basic.tex
pow2clk: I was confused by how that would differ from the using-declaration. I've consulted the C++ spec to see what that represents. It's not defined in the hlsl-specs text, but I expect it will eventually be here.

spec/basic.tex
pow2clk: I like the C++ language. I've proposed something possibly a little more succinct that splits the difference between the two and comes of clearer to me.

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+Every name that denotes a label is introduced by a \textit{labeled statement}
+(\ref{Stmt.Label}).
+
+\p A \textit{variable} is introduced by the declaration of a reference other
+than a non-static data member of an object. The variable's name denotes the
+reference or object.
+
+\p Whenever a name is encountered it is necessary to determine if the name
+denotes an entity that is a type or template. The process for determining if a
+name refers to a type or template is called \textit{name lookup}.
```
pow2clk:
```suggestion
\p A name can denote an entity that is a type or template. Whenever a name is
encountered it is necessary to determine if the name denotes such an entity.
The process for determining if a name refers to a type or template is called
\textit{name lookup}.
```

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+\begin{itemize}
+\item they are identifiers comprised of the same character sequence, or
+\item they are operator-function-ids formed with the same operator, or
+\item they are conversion-function-ids formed with the same type, or
+\item they are template-ids that refer to the same class or function.
+\end{itemize}
+
+\begin{note}
+  This section matches \gls{isoCPP} section \textbf{[basic]} except for the
+  exclusion of \texttt{goto} and \textit{literal operators}.
```
pow2clk: It doesn't match https://timsong-cpp.github.io/cppwp/n3337/basic#def exactly even with these exceptions. I don't actually agree with the decision to remove mention of gotos above, but it would seem consistent to remove it here as well. I think it is helpful to define HLSL at least in part in comparison to C++. We can write a spec that avoids that, but I think it would be harder to read.

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+without the \texttt{static} storage specifier has \textit{automatic storage
+duration}. Such an object is created when the block in which it is declared is
+entered and destroyed when the block is exited.
+
+\Sub{Program Storage Duration}{Basic.Storage.Program}
+
+\p An object whose name is declared in a global, namespace, or cbuffer scope
+without the \texttt{static} storage specifier has \textit{program storage
+duration}. Such an object is created when the program begins execution and
+destroyed when the program ends execution.
```
pow2clk, line 154: I don't think the use of "program" here matches what is described below as that encompasses the full result of the shader after compilation and linking. Particularly when contrasted with the description of thread group execution below, it might seem that the duration of this storage is greater than or equal to that of the thread group.

I realize it's not the same as the c++ definition of thread scope, but using the HLSL definition of "thread" or "lane" might be less confusing. If the first wave of lanes finishes, then this "program" has finished for the purposes of storage duration, but the "program" meaning the compiled and linked translation units, may still have further lanes to complete before it is done being executed.

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+inside the program begins.
+
+\p Entry point functions have different requirements based on the target runtime
+and execution mode (\ref{Basic.Start.Mode}).
+
+\p Parameters to entry functions and entry function return types must be of
+scalar, vector, or non-intangible class type (\ref{Basic.types}). Scalar and
+vector parameters and return types must be annotated with semantic annotations
+(\ref{Decl.Attr.Semantic}). Class type input and output parameters must have all
+fields annotated with semantic annotations.
```
pow2clk: Sometimes these can be denoted with qualifiers and in the case of mesh and geometry shaders or pseudo-templates rather than semantics. In some of those same cases, in addition to scalars and vectors, arrays thereof are sometimes required. Matrices are also viable in some stages.

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+
+\Sub{Execution Mode}{Basic.Start.Mode}
+
+\p A runtime may define a set of execution modes in an implementation defined
+way. Each execution mode will have a set of implementation defined rules which
```
pow2clk: I don't feel strongly whether we need a dash here or not, but there is some inconsistency in its usage as an adjective and I think we should pick one and be consistent. I have a very slight preference for a dash.
```suggestion
\p A runtime may define a set of execution modes in an implementation-defined
way. Each execution mode will have a set of implementation-defined rules which
```

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+\Sec{Types}{Basic.types}
+
+\p The \textit{object representation} of an object of type \texttt{T} is the
+sequence of \textit{N} bytes taken up by the object of type \texttt{T}, where
+\textit{N} equals \texttt{sizeof(T)}. The \textit{object representation} of an
+object may be different based on the \textit{memory space} it is stored in
+(\ref{Intro.Memory.Spaces}).
+
+\begin{note}
+  \texttt{sizeof(T)} returns thesize of the object as-if it's stored in device
```
pow2clk: Typos? The as-if usage is unfamiliar to me.
```suggestion
  \texttt{sizeof(T)} returns the size of the object as if it's stored in device
```

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+\texttt{T}, or if the type is known to be an intangible type. An object may not
+be defined to have an \textit{incomplete} type.
+
+\p Arithmetic types (\ref{Basic.types.arithmetic}), enumeration types, and
+\textit{cv-qualified} versions of these types are collectively called
+\textit{scalar types}.
+
+\p Vectors of scalar types declared with the built-in \texttt{vector<T,N>}
+template are \textit{vector types}. Vector lengths must be between 1 and 4 (i.e.
+\( 1 \leq N \leq 4 \) ).
```
pow2clk: Not anymore! Since it's a shader model restriction, I don't think it belongs in a language spec

spec/basic.tex
```diff
@@ -0,0 +1,475 @@
...
+The standard signed integer types and standard unsigned integer types are
+collectively called \textit{standard integer types}. The signed integer types
+and unsigned integer types are collectively called \textit{integer types}.
+Integer types inherit the object representation of integers defined in
+\glsdesc{isoC23}, which requires twos' complement representation. Integer types
+shall satisfy the constraints defined in \glsdesc{isoCPP}, section
+\textbf{basic.fundamental}.
+
+\p There are three \textit{standard floating point types}: \texttt{half},
+\texttt{float}, and \texttt{double}. The \texttt{float} type is a 32-bit
```
pow2clk, line 387: FWIW, float16_t, float32_t, and float64_t are also in the language and may deserve mention as the signed integer types above are.

spec/basic.tex
llvm-beanz: I think that whole note can go. I'm not sure it really adds relevant context for implementers. Referring to the C++ specification over and over again becomes difficult to maintain as the C++ specification changes over time, and it makes it so that our specification cannot stand on its own.

spec/basic.tex
llvm-beanz: I'm not sure what you're trying to say here. If you're referring to the `in`, `out` and `inout` keywords those aren't qualifiers (qualifiers are things like `const` which qualify the type).

C# calls those keywords "parameter modifiers" (https://ecma-international.org/wp-content/uploads/ECMA-334_7th_edition_december_2023.pdf), which is consistent with what we call them in the Clang and DXC implementations.

Irrespective, I think that signature validation varies pretty wildly based on the target runtime, so I don't think we can actually standardize that. The omission of matrix types I'll fix, I'm not sure what else you're looking for here.

spec/basic.tex
llvm-beanz: Ecma house style uses hyphens.

spec/basic.tex
llvm-beanz: I'm not sure what you mean by "after compilation and linking", storage duration is about the _lifetime_ of objects at runtime, not directly about compilation.

The duration of "program storage" is really the life of a dispatch. It's really about program-level constants (in the $Globals cbuffer), and resource declarations. It should be greater than the duration of a single thread group's storage.

spec/basic.tex
llvm-beanz: I tried an alternative simplification. Let me know if it works for you.

spec/basic.tex
llvm-beanz: I hope that I've clarified this wording.

spec/basic.tex
pow2clk: Yeah I see that now. What I referenced before was about the text below, but having thought about this more, I see how it can be conceived to have that lifetime. I no longer have any issue here.

spec/basic.tex
```diff
@@ -0,0 +1,478 @@
...
+than a non-static data member of an object. The variable's name denotes the
+reference or object.
+
+\p \textit{Name lookup} is the process of determining if a name refers to a type
+or template, and if so, which type or template it refers to. Name lookup occurs
+any time a name is encountered during parsing of a translation unit.
+
+\p Two names are the same name if:
+\begin{itemize}
+\item they are identifiers comprised of the same character sequence, or
```
jessebarker: Super nit: this should be "composed of" or "comprising".

spec/basic.tex
```diff
@@ -0,0 +1,478 @@
...
+\p Scalarized representations are determined as follows:
+\begin{itemize}
+\item The scalarized representation of an array \texttt{T[n]} is \(SR(T_0), ..
+SR(T_n)\).
+
+\item The scalarized representation of a vector \texttt{vector<T,n>} is \(T_0,
+.. T_n\).
+
+\item The scalarized representation of a matrix \texttt{matrix<T,n, m>} is
+\(T_0, .. T_{n \times m}\).
```
jessebarker: If you're doing 0-based representation here, shouldn't the final value be `n-1`?

jessebarker (approved):
LGTM

Feel free to ignore my nit(s).

spec/basic.tex
pow2clk: This says that inputs to entry functions "must be annotated with semantic annotations" and only lists scalars, vectors, and structs as options. Some entry points do not require semantic annotations and more types than those listed are allowed in some stages.

Mesh and geometry shader entry points have parameters whose purpose is indicated by qualifiers such as `payload`, `indices`, `vertices`, etc and have no semantic annotations (see mesh spec)[https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#vertex-indices]. Geometry shaders have similar. 

Hull shaders have parameters that indicate the meaning of the entry point input parameter with what I'd call a pseudo-template called (InputPatch)[with https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-inputpatch] which includes the type and array size in the template parameters.

Shader stages can take matrix inputs and tessellation and geometry stages require array types.

spec/basic.tex
pow2clk: It clarifies it a lot and answers a question I was going to ask about whether we were going to assume any common aspects of language without explicitly defining them.

spec/basic.tex
pow2clk: This eliminates the potential confusion that concerned me

spec/basic.tex
pow2clk: I'm gratified to see that the intent is to define everything even if it's redundant. Nevertheless, I think notes are a place where we can give asides from the most formal and parsimonious of language to deliver clarifying asides that might answer questions those familiar with c++ and particularly the c++ spec whose eyes might be glazing over with so much familiarity to the point that they miss the differences.

While I don't want to insist that this note specifically remain if I'm against consensus, I would like to have a conversation about what sorts of notes are worth mentioning. Is this one unneeded because a careful reading of the spec would make it clear there is no goto or is it because few people would assume that goto would find its way into a language designed after 1980?

spec/basic.tex
pow2clk: I see that you did respond to the different types feedback. I don't know why, but github is being completely useless in lining up old comments to updated text, making it hard for me to respond to these in light of that new text.

spec/basic.tex
pow2clk: I was using "dash" more casually than I should have. I really meant hyphens. Glad to see them added and to get that style info

spec/basic.tex
```diff
@@ -0,0 +1,478 @@
...
+\texttt{T}, or if the type is known to be an intangible type. An object may not
+be defined to have an \textit{incomplete} type.
+
+\p Arithmetic types (\ref{Basic.Types.Arithmetic}), enumeration types, and
+\textit{cv-qualified} versions of these types are collectively called
+\textit{scalar types}.
+
+\p Vectors of scalar types declared with the built-in \texttt{vector<T,N>}
+template are \textit{vector types}. An implementation must support vector
+lengths must be between 1 and 4 (i.e. \( 1 \leq N \leq 4 \) ).
```
pow2clk: I see the subtle change here, but I think it bears mention that implementations may support longer lengths

spec/basic.tex
pow2clk: Since this comment no longer shows among the changes, I've created a new one to address further concerns

spec/basic.tex
llvm-beanz: Is the note directly below this line not sufficient?

spec/basic.tex
jessebarker: (or maybe this allows for both 0-based and 1-based representations?)

pow2clk (approved):

llvm-beanz:

PR #24, opened 2026-04-27
Add Conversions Clause

llvm-beanz:
This is a mostly complete draft of the Conversions clause covering all the built-in conversions supported by HLSL. These conversions may occur implicitly or explicitly as will be described in other clauses where we describe expressions and overloading.


spec/conversions.tex
```diff
@@ -0,0 +1,295 @@
...
+
+\begin{note}
+  Array-to-pointer conversion of constant sized arrays is not supported.
+\end{note}
+
+\Sec{Integral promotion}{Conv.ipromote}
+
+\p An integral promotion is a conversion of:
+\begin{itemize}
+  \item a glvalue of integer type other than \texttt{bool} to a cxvalue of
```
pow2clk, line 58: The definition of cxvalue got left out of the basics section and as far as I can tell, isn't a standard term.

spec/conversions.tex
```diff
@@ -0,0 +1,295 @@
...
+\item The length of \texttt{A$^\prime$}, \texttt{N}, is greater than or equal to
+the length of \texttt{B$^\prime$}, \texttt{M}.
+\item For every index \texttt{I} from \texttt{0} to \texttt{M-1}, there exists
+an implicit conversion from \texttt{A$^\prime$[I]} to \texttt{B$^\prime$[I]}.
+\end{itemize}
+
+\p A glvalue of type \texttt{T[N]} can be converted to a cxvalue of type
+\texttt{T}.
+
+%{This is true if N is >= M but if N = M then there isn't really a cast}
```
pow2clk: Should this be a note?
```suggestion
\begin{note}
  This is true if N is >= M but if N = M then there isn't really a cast
\end{note}
```

spec/conversions.tex
```diff
@@ -0,0 +1,295 @@
+\Ch{Standard Conversions}{Conv}
+
+\p \acrshort{hlsl} inherits standard conversions similar to \gls{isoCPP}. This
+chapter enumerates the full set of conversions. A \textit{standard conversion
+sequence} is a sequence of standard conversions in the following order:
+\begin{enumerate}
+  \item Zero or one conversion of either lvalue-to-rvalue, or array-to-pointer.
+  \item Zero or one conversion of either integral conversion, floating point
+  conversion, floating point-integral conversion, or boolean conversion,
```
jessebarker: nit: does this "or" belong here?

spec/conversions.tex
```diff
@@ -0,0 +1,295 @@
...
+the values of the source type.
+
+\Sec{Floating point promotion}{Conv.fppromote}
+
+\p A glvalue of a floating point type can be converted to a cxvalue of a
+floating point type of higher conversion rank, or a prvalue of a floating point
+type can be converted to a prvalue of a floating point type of higher conversion
+rank.
+
+\p Floating point conversion ranks are defined in section \ref{Conf.rank.float}.
```
jessebarker: Should be `Conv.rank.float`.  Yields an unresolved section number reference ("??").

spec/conversions.tex
```diff
@@ -0,0 +1,295 @@
...
+be converted to a cxvalue of type \texttt{matrix<T,x,y>} or a prvalue of type
+\texttt{T} can be converted to a prvalue of type \texttt{matrix<T,x,y>}. The
+destination value is the source value replicated into each element of the
+destination.
+
+\Sec{Array and Class splat conversion}{Conv.asplat}
+
+\p A prvalue of type \texttt{T} can be converted to a prvalue of type
+\texttt{struct S}, if there are valid conversions from \texttt{T} to each
+\texttt{U} in flattened \texttt{S}, see section \ref{Decl.init.Agg} for
```
jessebarker: Are reference tags case sensitive?  This is `Decl.Init.Agg` both elsewhere in this doc, and in the declarations clause PR.

jessebarker (approved):
A couple of small things, but looks good otherwise.

spec/conversions.tex
llvm-beanz: Updated in the Basic Concepts PR.

spec/conversions.tex
llvm-beanz: I think that comment is redundant based on the spec language.

spec/conversions.tex
llvm-beanz: Yea, I'm going to do a pass to adopt consistent capitalization across all the references so that we have one consistent CamelCase style to help avoid this in the future.

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+\texttt{vector<T,M>}, if \texttt{N} is greater than or equal to \texttt{M}.
+
+\p A glvalue of type \texttt{vector<T,N>} can be converted to a cxvalue of type
+\texttt{T[M]}, if \texttt{N} is greater than or equal to \texttt{M}.
+
+\Sec{Vector splat conversion}{Conv.vsplat}
+
+\p A glvalue of type \texttt{T} can be converted to a cxvalue of type
+\texttt{vector<T,x>} or a prvalue of type \texttt{T} can be converted to a
+prvalue of type \texttt{vector<T,x>}. The destination value is the source value
```
pow2clk: It's a very minor nit, but for consistency with how vector lengths are indicated above, we may want to use N and M.
```suggestion
\texttt{vector<T,N>} or a prvalue of type \texttt{T} can be converted to a
prvalue of type \texttt{vector<T,N>}. The destination value is the source value
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+replicated into each element of the destination.
+
+\p A prvalue of type \texttt{vector<T,1>} can be converted to a prvalue of type
+\texttt{vector<T,x>}. The destination value is the value in the source vector
```
pow2clk: continued 
```suggestion
\texttt{vector<T,N>}. The destination value is the value in the source vector
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+replicated into each element of the destination.
+
+\Sec{Matrix splat conversion}{Conv.msplat} \p A glvalue of type \texttt{T} can
+be converted to a cxvalue of type \texttt{matrix<T,x,y>} or a prvalue of type
+\texttt{T} can be converted to a prvalue of type \texttt{matrix<T,x,y>}. The
```
pow2clk:
```suggestion
be converted to a cxvalue of type \texttt{matrix<T,M,N>} or a prvalue of type
\texttt{T} can be converted to a prvalue of type \texttt{matrix<T,M,N>}. The
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+\p A prvalue of type \texttt{vector<T,1>} can be converted to a prvalue of type
+\texttt{U[N]}, if there are valid conversions from \texttt{T} to each \texttt{V}
+in flattened \texttt{U}. The destination value is the value in the source vector
+replicated into each element of the flattened \texttt{U[N]}.
+
+\Sec{Vector and matrix truncation conversion}{Conv.VTrunc}
+
+\p A prvalue of type \texttt{vector<T,x>} can be converted to a prvalue of type:
+\begin{itemize}
+\item \texttt{vector<T,y>} only if \( y < x \), or
```
pow2clk: This gets more complicated of course. In keeping with the use of a prime symbol to indicate an altered form, perhaps we could use it on the dimensions? 

Left off making explicit change suggestions here due to uncertainty
```suggestion
\p A prvalue of type \texttt{vector<T,N>} can be converted to a prvalue of type:
\begin{itemize}
\item \texttt{vector<T,N$^\prime$>} only if \( N$^\prime$ < N \), or
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+
+\p Every integer and floating point type have defined conversion ranks. These
+conversion ranks are used to differentiate between \textit{promotions} and other
+conversions (see: \ref{Conv.IPromote} and \ref{Conv.FPPromote}).
+
+\Sub{Integer Conversion Rank}{Conv.Rank.Int}
+
+\begin{itemize}
+  \item  No two signed integer types shall have the same conversion rank even if
+  they have the same representation.
```
pow2clk, line 265: This refers to min precision types I expect? It doesn't apply to the likes of int32_t vs int does it? Those being aliases I suppose.

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+  of all standard integer types.
+  \item For all integer types \texttt{T1}, \texttt{T2}, and \texttt{T3}: if
+  \texttt{T1} has greater rank than \texttt{T2} and \texttt{T2} has greater rank
+  than \texttt{T3}, then \texttt{T1} shall have greater rank than \texttt{T3}.
+\end{itemize}
+
+\Sub{Floating Point Conversion Rank}{Conv.Rank.Float}
+
+\begin{itemize}
+  \item The rank \texttt{half} shall be greater than the rank of
```
pow2clk, line 285: I want to confirm that the rank of half doesn't change regardless of whether it is 16 bit or 32 bit, correct? I think in either case it should be above min16 and below float.

Similarly, I was about to ask how float16_t is positioned when half is 32 bits, but as it happens, that shouldn't be possible as float16_t is only allowed when half is sure to be 16-bit. I don't know if we want to assume that in the language spec though? If not, we may need to introduce float16_t and have conditions depending on whether half is 16 or 32 bit?

It's an unfortunate wrinkle

pow2clk (commented):
Some very minor nits you can take or leave and then an unfortunate 16 bit wrinkle I only just thought of.

spec/conversions.tex
llvm-beanz: It isn't strictly related to the min precision types. The C standard has similar language because the size of `short`, `int`, and `long` aren't defined by the standard, so it is a valid implementation of C if all three are 16-bit, but they still have different conversion ranks. Additionally `int` and `unsigned int` in C have the same object representation, but different ranks.

It comes into play for HLSL Both with the minimum precision types and signedness.

The explicit-sized types (e.g. `int32_t`, `uint16_t`,...) are type aliases of the base integer types not unique types, so they get the rank of the type they alias.

spec/conversions.tex
llvm-beanz: Yes, the rank of `half` is unchanged regardless of its size. `float16_t` is a type alias of `half` which is only defined when `half` is 16-bit. The language describing this is in #23.

pow2clk: I see the language. It is specifying the behavior of DXC more closely than I would have expected, but it sure does simplify things for now and since I hope we'll move further in the direction of always enabling 16-bit types, it's an ephemeral problem anyway.

llvm-beanz: DXC and Clang both follow that behavior. It's effectively a requirement for many targets that we want to be able to target with this standard.

DX did make 16-bit type support mandatory in Shader Model 6.9, but if we want standard HLSL to be portable to older runtimes we need to make 16-bit support optional and define how that behaves.

spec/conversions.tex
pow2clk: Since it only applies to signed integers, it wouldn't seem to relate to `int` versus `unsigned int` and their relations are stated below to be equal. 

I consulted the c++ spec to try to get better insight and was further confused by the exception listed: "No two signed integer types other than char and signed char (if char is signed) shall have the same rank, ..." I took that to mean that chars being an exception meant that `int` and `signed int` should have different ranks, but nowhere is it clear which should be higher. That also seems a strange rule. 

I can see how this is useful in the cases of types without defined sizes, so I really don't have any issue with this spec language though it only seems applicable to min16int in HLSL, but the rule proved by the exception in the original text makes me think I may not understand this properly

Since types are not aliased in the original spec as they are in HLSL, it would be useful to clarify that they don't apply to this here

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+\Sec{Vector and matrix truncation conversion}{Conv.VTrunc}
+
+\p A prvalue of type \texttt{vector<T,N>} can be converted to a prvalue of type:
+\begin{itemize}
+\item \texttt{vector<T,N$^\prime$>} only if \( N$^\prime$ < N \), or
+\item \texttt{T}
+\end{itemize}
+
+\p The resulting value of vector truncation is comprised of elements \( [0..y)
+\), dropping elements \( [y..x) \).
```
pow2clk: Encouraged by the acceptance of my previous prime proposals, I'll finish the job. 
```suggestion
\p The resulting value of vector truncation is comprised of elements \( [0..N$^\prime$)
\), dropping elements \( [N$^\prime$..N) \).
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+
+\p A prvalue of type \texttt{matrix<T,x,y>} can be converted to a prvalue of
+type:
+\begin{itemize}
+  \item \texttt{matrix<T,z,w>} only if \( x \geq z \) and \(y \geq w \),
```
pow2clk:
```suggestion
\p A prvalue of type \texttt{matrix<T,M,N>} can be converted to a prvalue of
type:
\begin{itemize}
  \item \texttt{matrix<T,M$^\prime$,N$^\prime$>} only if \( M \geq M$^\prime$ \) and \(N \geq N$^\prime$ \),
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+  \item \texttt{vector<T,z>} only if \( x \geq z \), or
```
pow2clk, line 219: I actually experimented with this and as far as I can tell, you can't cast matrices to vectors, a subscript is needed: https://godbolt.org/z/3exKoax9s

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+  \item \texttt{T}.
+\end{itemize}
+
+\p Matrix truncation is performed on each row and column dimension separately.
+The resulting value is comprised of vectors \( [0..z) \) which are each
+separately comprised of elements \( [0..w) \). Trailing vectors and elements are
```
pow2clk:
```suggestion
The resulting value is comprised of vectors \( [0..M$^\prime$) \) which are each
separately comprised of elements \( [0..N$^\prime$) \). Trailing vectors and elements are
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+dropped.
+
+\p Reducing the dimension of a vector to one (\texttt{vector<T,1>}), can produce
+either a single element vector or a scalar of type \texttt{T}. Reducing the rows
+of a matrix to one (\texttt{matrix<T,x,1>}), can produce either a single row
+matrix, a vector of type \texttt{vector<T,x>}, or a scalar of type \texttt{T}.
```
pow2clk:
```suggestion
of a matrix to one (\texttt{matrix<T,M,1>}), can produce either a single row
matrix, a vector of type \texttt{vector<T,M>}, or a scalar of type \texttt{T}.
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+
+\Sec{Component-wise conversions}{Conv.CWise}
+
+\p A glvalue of type \texttt{vector<T,x>} can be converted to a cxvalue of type
+\texttt{vector<V,x>}, or a prvalue of type \texttt{vector<T,x>} can be converted
+to a prvalue of type \texttt{vector<V,x>}. The source value is converted by
```
pow2clk:
```suggestion
\p A glvalue of type \texttt{vector<T,M>} can be converted to a cxvalue of type
\texttt{vector<V,M>}, or a prvalue of type \texttt{vector<T,M>} can be converted
to a prvalue of type \texttt{vector<V,M>}. The source value is converted by
```

spec/conversions.tex
```diff
@@ -0,0 +1,294 @@
...
+performing the appropriate conversion of each element of type \texttt{T} to an
+element of type \texttt{V} following the rules for standard conversions in
+chapter \ref{Conv}.
+
+\p A glvalue of type \texttt{matrix<T,x,y>} can be converted to a cxvalue of
+type \texttt{matrix<V,x,y>}, or a prvalue of type \texttt{matrix<T,x,y>} can be
+converted to a prvalue of type \texttt{matrix<V,x,y>}. The source value is
```
pow2clk:
```suggestion
\p A glvalue of type \texttt{matrix<T,M,N>} can be converted to a cxvalue of
type \texttt{matrix<V,M,N>}, or a prvalue of type \texttt{matrix<T,M,N>} can be
converted to a prvalue of type \texttt{matrix<V,M,N>}. The source value is
```

pow2clk (commented):
Mostly finishing up the size notation conversion I started earlier, but I did note something when I tested matrix conversion to vector since it was unfamiliar to me and DXC doesn't accept it.

Interested in further clarification of how the signed integer rank language applies to HLSL, but I don't actually think it's harmful now, but would like to see aliases addressed and confirm that I've understood it correctly

spec/conversions.tex
llvm-beanz: By "original spec" I assume you mean the C++ spec. There is no original HLSL spec. I would caution against reading small bits of the C++ spec in isolation. This will also be true for the HLSL spec, but these languages are _so complicated_ that you can't just read one subclause and understand behavior.

In C++, the explicitly sized integer types are type aliases. The difference with HLSL is that C++ defines the alias in terms of the standard library and which header they're defined in (https://timsong-cpp.github.io/cppwp/n3337/cstdint#syn).

DXC doesn't have a real notion of standard library headers (although that has been something slowly changing), so in the draft HLSL specification I just denoted them as aliases in the basic types section instead of defining them as such in the "standard library" definition.

The C++ language requires a bit of additional context.

Consider this:

```c++
#include <type_traits>

static_assert(std::is_same<int, signed>::value, "woo!");
static_assert(std::is_same<int, signed int>::value, "woo!");
```
[Compiler Explorer](https://godbolt.org/z/veGa6bcKM)

These static_asserts both evaluate to true.

To understand why, you need the context in the [decl.type](https://timsong-cpp.github.io/cppwp/n3337/dcl.type) clause. Specifically the table in [decl.type.simple](https://timsong-cpp.github.io/cppwp/n3337/dcl.type#simple) provides a mapping of simple type specifiers to their types.

C++ deviates from C here in a few key ways, like the fact that `int` is defined to be signed so `signed int` and `int` are the same type (in C that is also implementation defined as it is for `char`).

Do you have concrete suggestions for how the wording in this section should change?

spec/conversions.tex
llvm-beanz: This is a place where DXC has some strangely inconsistent rules. You can't cast a float4x4 to a float4, but you can cast it to a struct containing nothing but a float4... then cast that to a float4.

```hlsl
struct S {
    float4 F;
};

[numthreads(1,1,1)]
void main()
{
    float4x4 mat = 0;
    float4 f = (float4)(S)mat;
}
```
[Compiler Explorer](https://godbolt.org/z/1bs11fPv3)

The rules we settled on in the spec here apply consistency, rather than preserving inconsistent special handling of types.

pow2clk (approved):

llvm-beanz:

PR #35, opened 2026-04-30
Add initial expressions clause

llvm-beanz:
This adds some of the initial text for the expressions clause.

The main language defined here are the usual arithmetic conversions, vector and matrix member accessors, and function call behavior.


spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+
+\p Whenever a \textit{glvalue} appears in an expression that expects a
+\textit{prvalue}, a standard conversion sequence is applied based on the rules
+in \ref{Conv}.
+
+\Sec{Usual Arithmetic Conversions}{Expr.Conv}
+
+\p Binary operators for arithmetic and enumeration type require that both
+operands are of a common type. When the types do not match the \textit{usual
+arithmetic conversions} are applied to yield a common type. When \textit{usual
```
pow2clk: I will usually turn a blind eye to missing commas for dependent clauses, but this one ending in a verb with no object made it hard to read correctly. I first read not that the types were mismatching each other but that they were mismatching the `usual arithmetic conversions`, which clearly doesn't make sense, but that's how I parsed it. 
```suggestion
operands are of a common type. When the types do not match, the \textit{usual
arithmetic conversions} are applied to yield a common type. When \textit{usual
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+arithmetic conversions} are applied to vector operands they behave as
+component-wise conversions (\ref{Conv.CWise}). The \textit{usual arithmetic
+conversions} are:
+
+\begin{itemize}
+  \item If either operand is of scoped enumeration type no conversion is
```
pow2clk: This one reads okay without the comma, but the rest of this list uses them.
```suggestion
  \item If either operand is of scoped enumeration type, no conversion is
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+  X>}, the other operand shall be converted to match.
+  \item Otherwise, if either operand is of type \texttt{half} or \texttt{vector<half, X>},
+  the other operand shall be converted to match.
+  \item Otherwise, integer promotions are performed on each scalar or vector
+  operand following the appropriate scalar or component-wise conversion
+  (\ref{Conv}).
+  \begin{itemize}
+    \item If both operands are scalar or vector elements of signed or unsigned
+    types, the operand of lesser integer conversion rank shall be converted to
+    the type of the operand with greater rank.
```
pow2clk: Should this specify integer types?
```suggestion
    \item If both operands are scalar or vector elements of signed or unsigned
    integer types, the operand of lesser integer conversion rank shall be
    converted to the type of the operand with greater rank.
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+  primary-expression\br
+  % The [] characters on the two lines below should be inside \terminal, however
+  % pandoc doesn't seem to like that.
+  postfix-expression [ expression ]\br
+  postfix-expression [ braced-init-list ]\br %
+  postfix-expression \terminal{(} \opt{expression-list} \terminal{)}\br
+  simple-type-specifier \terminal{(} \opt{expression-list} \terminal{)}\br
+  typename-specifier \terminal{(} \opt{expression} \terminal{)}\br
+  simple-type-specifier braced-init-list\br
+  typename-specifier braced-init-list\br
```
pow2clk: Lingering effects syntax prevents us from supporting these methods of initialization
```suggestion
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+  postfix-expression \terminal{.} \opt{\terminal{template}} id-expression\br
+  postfix-expression \terminal{.} vector-swizzle-component-sequence\br
+  postfix-expression \terminal{.} matrix-swizzle-component-sequence\br
+  postfix-expression \terminal{++}\br
+  postfix-expression \terminal{--}
+\end{grammar}
+
+\Sec{Subscript}{Expr.Post.Subscript}
+
+% TODO: Language pending proposal in https://github.com/hlsl-tc57/tc57/pull/27.
```
pow2clk: As the PR has been merged, perhaps this should reference the proposal in the repo### 
```suggestion
% TODO: Language pending proposal in https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md.
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+\end{itemize}
+
+\p A \textit{matrix-swizzle-component-sequence} is a sequence of one but less
+than or equal to four swizzle components of either the
+\textit{matrix-zero-indexed-swizzle} or \textit{matrix-one-indexed-swizzle}
+forms; the two forms may not be mixed in the same
+\textit{matrix-swizzle-component-sequence}. The type of a swizzle expression is
+a vector of the same element type as the postfix expression before the dot,
+with a number of elements equal to the number of components in the
+\textit{matrix-swizzle-component-sequence}.
```
pow2clk, line 264: I just realized we might have technically expanded this when we introduced long vectors. Given how slightly odd it already is and how much odder it would be if you could chain up to 2048 swizzle components, I think I prefer we leave it at four.

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+expressions are evaluated before the function call occurs.
+
+\p \textit{Input parameters} are passed by-value into a function. If an argument
+to an \textit{input parameter} is of constant-sized array type, the array is
+copied to a temporary and the temporary value is converted to an address via
+array-to-pointer decay. If an argument is an unsized array type, the array
+lvalue directly decays via array-to-pointer decay.
+
+% TODO: The language above matches DXC's behavior but will need to be updated to
+% match the final specification from https://github.com/hlsl-tc57/tc57/pull/27.
```
pow2clk:
```suggestion
% match the final specification from https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md.
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+same rules for any other data type. If an argument to an \textit{output} or
+\textit{input/output parameter} is an unsized array type, the array lvalue
+directly decays via array-to-pointer decay. An argument of a constant sized
+array of type \texttt{T[N]} can be converted to a cxvalue of an unsized array
+of type \texttt{T[]} through array to pointer decay. An unsized array of type
+\texttt{T[]}, cannot be implicitly converted to a a constant sized array of type
+\texttt{T[N]}.
+
+% TODO: The language above matches DXC's behavior but will need to be updated to
+% match the final specification from https://github.com/hlsl-tc57/tc57/pull/27.
```
pow2clk:
```suggestion
% match the final specification from https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md.
```

spec/expressions.tex
```diff
@@ -0,0 +1,387 @@
...
+
+\p The value of a function call is the value returned by the called function.
+
+\p A function call is an lvalue if the result type is an lvalue reference type;
+otherwise it is a prvalue.
+
+\begin{note}
+  Functions that return lvalue references are unspellable in HLSL, but the
+  language requires them for some built-in functions. For example, the
+  \texttt{operator[]} overloads for writable resource types.
```
pow2clk, line 381: From a language-perspective, that's a different postfix expression. I don't know of any built-in functions that do this

spec/expressions.tex
llvm-beanz: These syntaxes aren't supported by any HLSL compiler today (although admittedly they only fail in Clang due to a bug), so I'm happy to remove them.

That aside, I'm not particularly familiar with the effects syntax can you elaborate on how that impacts here? These are examples of these formulations in code:

```c++
template <typename T>
struct S {
    using type = T; 

    void fn() {
      auto v = typename S<T>::type{2}; // typename-specifier braced-init-list
    }
};

[numthreads(1,1,1)]
void main() {
    auto f = float{1}; // simple-type-specifier braced-init-list
}
```

spec/expressions.tex
llvm-beanz: It isn't strictly a different postfix expression because it can be written of the form:

```hlsl
RWBuffer<float> In;

[numthreads(1,1,1)]
void main() {
    In.operator [](1);
}
```

I believe the other occurrence is the `Get` functions on the work graphs types.

pow2clk: The first is a pretty strange thing to do, but the `Get`s are the designated way to do things. Thanks for the insight

spec/expressions.tex
pow2clk: To be honest, I'm not familiar with effects syntax either except that it involved a lot of {} and that DXC complains of ignoring deprecated syntax when I tried to use them for initializers : https://godbolt.org/z/7rzs1r5eM

In the example you give, I believe it's creating temporary float values, initializing them, then assigning them to the autos. In this case, DXC instead confuses the syntax for being casts and fails as well

The introduction of `auto` might make such formulations more interesting and we should make a note to consider it along with the other 202x features, which, happily, includes removal of effects syntax.

pow2clk (approved):
Just added more context to my comments about postfix operations, both in resolved comments

spec/expressions.tex
llvm-beanz: I'm hesitant to incorporate any syntax that depends on braced-init-list before we fix braced initializers...

jessebarker (approved):
LGTM

llvm-beanz:

PR #37, opened 2026-05-11
Add initial declarations clause

llvm-beanz:
This adds the beginning of the declarations clause. The grammar specification is reasonably complete at the top heading, but the sub-clauses are mostly empty.

The initializers subclause contains spec language for HLSL's initialization list behavior used for aggregate initialization of all data types.


spec/declarations.tex
```diff
@@ -0,0 +1,132 @@
...
+  \define{special-declaration}\br
+  export-declaration-group\br
+  cbuffer-member-declaration\br
+
+  \define{block-declaration}\br
+  simple-declaration\br
+  namespace-alias-definition\br
+  using-declaration\br
+  using-directive\br
+  static\_assert-declaration\br
```
pow2clk, line 31: Do we have static_assert yet?

spec/declarations.tex
```diff
@@ -0,0 +1,132 @@
...
+  alias-declaration\br
+  opaque-enum-declaration\br
```
pow2clk, line 33: Will the undefined symbols come along with the later addition of the namespace section or should we import their definitions into this grammar?

spec/declarations.tex
```diff
@@ -0,0 +1,132 @@
...
+
+\end{grammar}
+
+\Sec{Specifiers}{Decl.Spec}
+\Sub{General}{Decl.Spec.General}
+\p The specifiers that can be used in a declaration are
+\begin{grammar}
+  \define{decl-specifier}\br
+  \textit{function-specifier}\br
+  ...
```
pow2clk: Does the ellipsis mean to be filled in later? HLSL supports `typedef` at least. At present it looks like the literal `...` can be parsed as a decl-specifier.

spec/declarations.tex
```diff
@@ -0,0 +1,132 @@
...
+\p The \texttt{export} specifier denotes that the function has program linkage (\ref{Basic.Linkage.Program}).
+
+\p The \texttt{export} specifier cannot be used on functions directly or indirectly within an unnamed namespace.
+
+\p Functions with program linkage can also be specified in \textit{export-declaration-group} (\ref{Decl.Export}).
+
+\p If a function is declared with an \texttt{export} specifier then all redeclarations of the same function must also use the \texttt{export} specifier or be part of \textit{export-declaration-group} (\ref{Decl.Export}).
+
+\Sec{Declarators}{Decl.Decl}
+% TODO: Complete the declarators subclause.
```
pow2clk, line 73: What is missing?

spec/declarations.tex
llvm-beanz: We do have C's `_Static_assert` in DXC (we use it for a bunch of testing). Adopting the C++11 (and maybe later) formations is proposed in #51. This part of the spec doesn't define which declaration we end up supporting.

spec/declarations.tex
llvm-beanz: My intent is to fill in the grammars as we write each section. Happy to discuss alternative approaches.

spec/declarations.tex
llvm-beanz: Updated to include the full grammar.

spec/declarations.tex
llvm-beanz: Not sure what you're asking here. There isn't anything "missing" the use of anonymous namespaces is generally a practice for making declarations visible only within a single translation unit, so the `export` keyword and unnamed namespaces are features that directly contradict each other.

pow2clk: The comment says the Declarators subclause needs completing, my question was what was incomplete. I see now that it is a completely empty section and not the section head of the below as subsections as I'd originally read it

This did make me wonder about the other missing sections that I expect will be needed that aren't included here as empty sections with "TODO" comments. I don't imagine we'll forget them either way, but I wonder if this one is being included special while the other pending sections are not for a reason

Perhaps that reason is the very fact that it is being included as a subclause of the declarations as it is in the latest C++ spec. That represents a slight deviation from most of the section organization as I've seen it in the rest of the spec that draws from c++11. It was something that occurred to me elsewhere in places that I liked the latest organization better. We can obviously do whatever makes most sense for HLSL, but I would like to know what is our overall approach here?

pow2clk (approved):
My only comment was to clarify what my admittedly overly terse statement meant, but also to introduce a subject I'd like to discuss regarding the overall organization strategy of the spec

jessebarker (approved):
LGTM

llvm-beanz:

PR #46, opened 2026-05-18
[Editorial] Add Annex A as grammar summary

llvm-beanz:
This adds Annex A as an informative annex to collect all the grammar productions from around the specification into a single reference.

### DIFF
--- a/spec/annex-a.tex
+++ b/spec/annex-a.tex
@@ -1,0 +1,7 @@
+\Annex{A}{informative}{Grammar Summary}{AnnexA}
+
+\closegrammarcollection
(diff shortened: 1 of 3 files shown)


llvm-beanz:
@hlsl-tc57/tc57-editors, this is just a change to LaTeX macros to automatically generate a grammar annex, and the basic annex header formatting to match Ecma house style.

Can I get a review from one of the editors on this?

bogner (approved):
It's a little bit awkward to spread state out like this, and importantly it only works if the annex is after all of the grammar collections, but I can't really come up with a way to do this that's any less complex. LGTM.

Alexander-Johnston (approved):

PR #47, opened 2026-05-18
[New] Add proposal for memory and execution model

llvm-beanz:
This PR adds a rough proposal for HLSL to define a memory and execution model. The proposal in its current state doesn't align on a single solution, instead it is written to consider the possible solution space and work through the options within the committee.

### DIFF
--- a/proposals/0007-mem-and-exec-model.md
+++ b/proposals/0007-mem-and-exec-model.md
@@ -152,7 +152,7 @@ Specific concerns that must be addressed:
 * What are the ordering requirements, if any, for memory operations to aliasing
   memory across a set of tangled threads?
 
-## Appendix 1: Magic Decoder Ring
+## Appendix 1: Glossary
 
 | DirectX Term | Khronos Term | Description |
 | ------------ | ------------ | ----------- |
--- a/proposals/0007-mem-and-exec-model.md
+++ b/proposals/0007-mem-and-exec-model.md
@@ -1,5 +1,5 @@
 ---
-title: NNNN - Formalized Memory and Execution Model
+title: 0007 - Formalized Memory and Execution Model
 params:
   authors:
   - llvm-beanz: Chris Bieneman
--- a/proposals/NNNN-mem-and-exec-model.md
+++ b/proposals/NNNN-mem-and-exec-model.md
@@ -49,7 +49,7 @@ of fall through.
 ### Proposed solution #1 : Full Lockstep
 
 A full lockstep execution model is the simplest to understand. Under full
-lockstep, all the threads in a warp must behave as if they share the same
+lockstep, all the threads in a wave must behave as if they share the same
 program counter whether they are in the same tangle or not.
 
 This execution model provides some of the strictest guarantees for memory
@@ -137,7 +137,7 @@ becomes unintuitive and potentially undefined.
 
 This can be made slightly stricter by requiring that branch statements (`if`,
 `else`, `switch`, `for`, `while`, etc.) are thread sync points. It also likely
-requires atomic operations to behave a sync points as well.
+requires atomic operations to behave as sync points as well.
 
 ## Memory Model Proposals
 
--- a/proposals/NNNN-mem-and-exec-model.md
+++ b/proposals/NNNN-mem-and-exec-model.md
@@ -1,0 +1,162 @@
+---
+title: NNNN - Formalized Memory and Execution Model
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+This proposal seeks to define the memory and execution models for HLSL. The goal
+is to define a memory and execution model that is understandable to users,
+portable across a wide variety of GPU hardware, and strikes a balance between
+full portability and performance.
+
+## Motivation
+
+The HLSL and DXIL memory and execution model have never been fully defined. This
+forces reliance on Windows device certification for behavior conformance, and a
+general approach that the implementation defines the behavior.
+
+This state is not ideal from the start, however as HLSL becomes more widely
+portable, and DirectX moves to SPIRV the lack of a documented memory and
+execution model makes it near impossible to ensure portability across byte code
+formats.
+
+The SPIRV-defined memory and execution model is constantly evolving and seeking
+to address some of the specific problems discussed in this proposal, however
+SPIRV is not designed to be written by humans. It is not a requirement that
+HLSL's memory and execution models match SPIRV, just that SPIRV is expressive
+enough to model programs in the model defined by HLSL.
+
+This proposal is structured with multiple proposed solutions. Those will become
+_Alternatives Considered_ as they are eliminated from consideration. There are
+also two groupings of proposals to capture the execution and memory models
+separately although they are tightly connected in the final representation.
+
+## Execution Model Proposals
+
+All execution model proposals assume behaviors documented in
+[SPV_KHR_maximal_reconvergence](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_maximal_reconvergence.asciidoc),
+as well as additional reconvergence requirements for `OpSwitch` to match the
+DXIL `switch` behavior such that tangles are formed by branch target (not
+selector value), and tangles are expected to reconverge at labels in the event
+of fall through.
+
+### Proposed solution #1 : Full Lockstep
+
+A full lockstep execution model is the simplest to understand. Under full
+lockstep, all the threads in a warp must behave as if they share the same
+program counter whether they are in the same tangle or not.
+
+This execution model provides some of the strictest guarantees for memory
+ordering and behavior. Consider the following code snippet:
+
+<a name="example1"></a>
+
+```hlsl
+groupshared int X;
+
+[numthreads(4,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  if (GI == 0)
+    X = 0;
+  else if (GI == 2)
+    X = 2;
+}
+```
+
+In full lockstep this program is well-defined. Because all threads must act
+as-if they share a program counter, no thread can execute the `else if`
+condition or its block until thread 0 has executed the body of the `if`. This
+enforces strict ordering of memory operations to `groupshared`.
+
+### Proposed solution #2 : Lockstep Within A Tangle
+
+The lockstep within a tangle execution model is a slightly relaxed variant of
+the full lockstep model. It allows each tangle to have an independent program
+counter. In this model the [example from solution #1](#example1) is undefined
+because once thread 0 splits to form its own tangle, the second tangle can
+continue executing until it reaches a reconvergence point. This creates a data
+race writing to X between thread 0 and 2.
+
+This model requires more precise definition of reconvergence points, however it
+provides some ordering guarantees that make it safe. For example, the following
+adjusted program is well-defined in this model.
+
+<a name="example2"></a>
+
+```hlsl
+groupshared int X;
+
+[numthreads(4,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  if (GI == 0)
+    X = 0;
+  if (GI == 2) // Reconverges here because this is a new statement, not an else.
+    X = 2;
+}
+```
+
+### Proposed solution #3 : Independent Threads
+
+This model is the most complicated, but also most flexible for compiler backend
+optimization. In this model threads are allowed to have fully independent
+program counters which only need to synchronize across tangles at designated
+sync points (e.g. wave operations, barriers, etc).
+
+Under this model the [example from solution #2](#example2) is undefined because
+the program contains no synchronization points, so each thread can execute
+independently and the memory ordering is not guaranteed.
+
+To illustrate a particular challenge with this model, consider the following
+example:
+
+<a name="example3"></a>
+
+```hlsl
+groupshared int X;
+
+[numthreads(4,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  if (GI == 0)
+    X = 0;
+  GroupMemoryBarrierWithGroupSync(); // sync point!
+  if (X == 0)
+    X = 2; // How many threads are in the tangle here?
+}
+```
+
+One problem with this model is not having clearly defined memory ordering which
+can impact tangle formation. If a thread is allowed to execute the second `if`
+body before all threads have finished evaluating the condition, tangle formation
+becomes unintuitive and potentially undefined.
+
+This can be made slightly stricter by requiring that branch statements (`if`,
+`else`, `switch`, `for`, `while`, etc.) are thread sync points. It also likely
+requires atomic operations to behave a sync points as well.
+
+## Memory Model Proposals
+
+Basically all modern GPUs implement some version of the [Heterogeneous Systems
+Architecture
+(HSA)](https://en.wikipedia.org/wiki/Heterogeneous_System_Architecture)
+standards. This makes it reasonable that HLSL's memory model derive from HSA.
+
+Specific concerns that must be addressed:
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a wave?
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a set of tangled threads?
+
+## Appendix 1: Magic Decoder Ring
+
+| DirectX Term | Khronos Term | Description |
+| ------------ | ------------ | ----------- |
+| thread, lane | invocation   | The computation performed on a single element as described in the program. |
+|              | tangle       | A grouping of co-executing threads. |
+| wave         | subgroup     | A group of threads which may form one or more tangles and are executed on a shared SIMD or other compute unit. |
+| threadgroup  | workgroup    | A group of threads which may be subdivided into one or more waves and comprise a larger computation. |
--- a/proposals/NNNN-mem-and-exec-model.md
+++ b/proposals/NNNN-mem-and-exec-model.md
@@ -22,15 +22,15 @@ forces reliance on Windows device certification for behavior conformance, and a
 general approach that the implementation defines the behavior.
 
 This state is not ideal from the start, however as HLSL becomes more widely
-portable, and DirectX moves to SPIRV the lack of a documented memory and
+portable and DirectX moves to SPIR-V, the lack of a documented memory and
 execution model makes it near impossible to ensure portability across byte code
 formats.
 
 The SPIRV-defined memory and execution model is constantly evolving and seeking
 to address some of the specific problems discussed in this proposal, however
-SPIRV is not designed to be written by humans. It is not a requirement that
-HLSL's memory and execution models match SPIRV, just that SPIRV is expressive
-enough to model programs in the model defined by HLSL.
+SPIR-V is not designed to be written by humans. It is not a requirement that
+HLSL's memory and execution models match SPIR-V, just that the HLSL model can be
+expressed using SPIR-V.
 
 This proposal is structured with multiple proposed solutions. Those will become
 _Alternatives Considered_ as they are eliminated from consideration. There are


proposals/NNNN-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+One problem with this model is not having clearly defined memory ordering which
+can impact tangle formation. If a thread is allowed to execute the second `if`
+body before all threads have finished evaluating the condition, tangle formation
+becomes unintuitive and potentially undefined.
+
+This can be made slightly stricter by requiring that branch statements (`if`,
+`else`, `switch`, `for`, `while`, etc.) are thread sync points. It also likely
+requires atomic operations to behave a sync points as well.
```
devshgraphicsprogramming: We prefer this, or Vulkan perf will be nerfed.

Sure we've ran into problem with it before 
https://graphics-programming.org/blog/subgroup-shuffle-execution-dependency-on-nvidia

But this is why SPIR-V gives you a subgroup execution barrier which you can liberally pepper around

proposals/0007-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+Basically all modern GPUs implement some version of the [Heterogeneous Systems
+Architecture
+(HSA)](https://en.wikipedia.org/wiki/Heterogeneous_System_Architecture)
+standards. This makes it reasonable that HLSL's memory model derive from HSA.
+
+Specific concerns that must be addressed:
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a wave?
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a set of tangled threads?
```
devshgraphicsprogramming, line 153: Vulkan has a pretty neat memory model for this which plays nicely with SPIR-V VMM.

Essentially its like C++11 with the addition of agent scopes Device, Workgroup, Subgroup, RayInvocation etc.

so if you can already reason about C++11 availability/visibility you're just adding the venn diagrams of the different scopes into the mix. C++11 memory model is basically similar to VMM except there's only a single scope.

I'd want to see HLSL Vulkan memory specced against VMM (or at  least C++11), so I know what variable reads/writes do as OpLoad/OpStore/OpAtomicXXX w.r.t. the domain and acquire/release

devshgraphicsprogramming: Also in VMM there's this feature/property called Availability Visibility Chains, its mighty useful unfortunately many old AMD GPUs and Apple GPUs don't support it.

proposals/NNNN-mem-and-exec-model.md
Tobski: Making branch statements implicitly have the equivalent of a Vulkan subgroup control barrier makes sense to me, since that's more or less the time you want to use it to avoid weird interactions with hoisting letting memory accesses escape the tangle.

I'm not clear why you'd want atomics to be a sync point though - what's the motivation for this?

proposals/NNNN-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+[SPV_KHR_maximal_reconvergence](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_maximal_reconvergence.asciidoc),
+as well as additional reconvergence requirements for `OpSwitch` to match the
+DXIL `switch` behavior such that tangles are formed by branch target (not
+selector value), and tangles are expected to reconverge at labels in the event
+of fall through.
+
+### Proposed solution #1 : Full Lockstep
+
+A full lockstep execution model is the simplest to understand. Under full
+lockstep, all the threads in a warp must behave as if they share the same
```
Tobski: nit: s/warp/wave?

proposals/0007-mem-and-exec-model.md
Tobski: Vulkans' av/vis chains are a little hard to parse but yea that is the Vulkan answer to the "across set of threads in a tangle/wave" part.

However, I wasn't sure if these questions were also asking about aliasing in general since that's also kind of a whole other thing with the memory model. IIRC HLSL basically bans most forms of resource aliasing; whereas Vulkan allows it with some caveats. The VMM answer for that part is the idea of references vs. memory locations. The memory location is the underlying memory being written to, where a reference is something like a descriptor or a virtual address. If you alias the same memory via different references, there's no guarantee that you see writes to one through reads to the other, ever. If the references _and_ memory locations line up, then you get the writes to one alias showing up via the other one if you either memory barrier them or tell the compiler they're an alias. Coordinating that across threads isn't too hard.

Tobski (commented):
Looks like a solid start, I'm definitely interested to see model 3 progress in terms of how branches are specified.

proposals/NNNN-mem-and-exec-model.md
devshgraphicsprogramming: whats the  opt out meachism from getting implicit subgroup control barrier and maximal reconvergence on all shaders and not have my Nvidia performance get nerfed due  to not being able to use IPC ?

> I'm not clear why you'd want atomics to be a sync point though - what's the motivation for this?

If you're referencing our blogpost it was an NV compiler bug.

AFAIK subgroup shuffle shuffles registers around, and due to IPC register values werent in lockstep. It wasnt  the atomic that needed to be barriered, it was the register its return value was being stored to, if that makes sense.

Also the only thing you'd need is an execution barrier but for some reason just a memory barrier fixed it too, but that just might be due to how NV compiler and HW works.

proposals/0007-mem-and-exec-model.md
devshgraphicsprogramming: >  IIRC HLSL basically bans most forms of resource aliasing; whereas Vulkan allows it with some caveats.

We went into it in great detail during the Shading Language Symposium, basically Shader SPIR-V bans it for everything thats NOT backed by VkDeviceMemory or groupshared, and for aliasing to work on groupshared you need a bunch of extensions.
(The Storage Classes where as a programmer you can't reasonably reason about the memory location of an OpVariable)

https://www.khronos.org/assets/uploads/developers/presentations/Francisco-Letterio-DevSH_SPIRV.pdf

For all other things OpVariables of descriptors can be decorated as Aliased, AliasedPointer can be applied to BDA but its not baked into the type (although that has its own set of problems https://github.com/KhronosGroup/SPIRV-Registry/issues/388 ).

But I'm preaching to the choir here, you already know all this.

> via the other one if you either memory barrier them or tell the compiler they're an alias

Weird I would have thought you need both...

Does this mean we've finally found the one single usecase of an `OpMemoryBarrier` with `Memory` Scope being Invocation?

> Vulkans' av/vis chains are a little hard to parse but yea that is the Vulkan answer to the "across set of threads in a tangle/wave" part.

I was under the impression that AV-Vis chains were for, "I can form Avail-Vis pair via transitive chain of execution dependencies".

What did I get wrong?

proposals/0007-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+The HLSL and DXIL memory and execution model have never been fully defined. This
+forces reliance on Windows device certification for behavior conformance, and a
+general approach that the implementation defines the behavior.
+
+This state is not ideal from the start, however as HLSL becomes more widely
+portable, and DirectX moves to SPIRV the lack of a documented memory and
+execution model makes it near impossible to ensure portability across byte code
+formats.
+
+The SPIRV-defined memory and execution model is constantly evolving and seeking
```
amd-mherdeg, line 29: SPIR-V supports multiple memory models and execution models:
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Memory_Model
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Execution_Model

When you say "the SPIRV-defined memory and execution model", are you referring to OpenCL, which confusingly does use that terminology?
https://registry.khronos.org/SPIR-V/specs/1.0/SPIR-V-execution-and-memory-model.pdf

More relevant are the Vulkan memory and execution models:
https://docs.vulkan.org/spec/latest/appendices/memorymodel.html
https://docs.vulkan.org/spec/latest/chapters/fundamentals.html#fundamentals-execmodel

The Vulkan memory model uses a SPIR-V extension (https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_vulkan_memory_model.html).

Even though this is an HLSL proposal, aren't you really formalizing the D3D memory and execution model, which then maps to features in both HLSL and SPIR-V, including probably a SPIR-V extension?

llvm-beanz: > Even though this is an HLSL proposal, aren't you really formalizing the D3D memory and execution model, which then maps to features in both HLSL and SPIR-V, including probably a SPIR-V extension?

No, this is about formalizing the memory and execution model for HLSL not a runtime. The memory model for HLSL does not need to match the memory model of the target runtime, we just need sufficient primitives in the target ISAs to map HLSL's model to the target.

The goal of defining a language-level memory and execution model is to give predictability and portability to HLSL code regardless of what runtime it runs on, and we will drive necessary changes into SPIRV and DXIL to facilitate this.

proposals/NNNN-mem-and-exec-model.md
Tobski: > I'm not clear why you'd want atomics to be a sync point though - what's the motivation for this?

This was a question about the assertion in the proposal that this might be needed.

Tobski: > whats the opt out meachism from getting implicit subgroup control barrier and maximal reconvergence on all shaders and not have my Nvidia performance get nerfed due to not being able to use IPC ?

Barriers that don't actually order anything can be elided. A barrier at the branch point only really matters for the purpose of making sure you don't get stuff jumping out of the branch. If a branch doesn't include a relevant memory access, the HLSL compiler should omit the barrier.

In the SPIR-V memory model, a "relevant access" would any "non-private" access; similarly in HLSL this would be any atomic or coherent access only.

proposals/0007-mem-and-exec-model.md
Tobski: > Does this mean we've finally found the one single usecase of an OpMemoryBarrier with Memory Scope being Invocation?

I mean, I've known about it for a long time 😅. It's a slight annoyance to me how badly aliasing is integrated into the memory model at present precisely because of this. There's probably a world in which you could imagine a front end compiler never omitting aliasing info; instead choosing to express the constraints solely with memory barriers. But we would need tighter barriers (on the order of affecting specific accesses) before that could be a practical reality.

> I was under the impression that AV-Vis chains were for, "I can form Avail-Vis pair via transitive chain of execution dependencies".
> 
> What did I get wrong?

Nothing, GitHub placed this comment weirdly, it's meant to be agreeing with you on that side of things though 😅

Tobski: > There's probably a world in which you could imagine a front end compiler never omitting aliasing info; instead choosing to express the constraints solely with memory barriers. But we would need tighter barriers (on the order of affecting specific accesses) before that could be a practical reality.

...I should probably raise an internal issue about this.

Tobski: Worth noting that right now using a memory scope of Invocation probably _won't_ guarantee this, because AFAIU it's not tested for this. Control barriers even say they ignore memory semantics if the Memory scope is Invocation.

proposals/NNNN-mem-and-exec-model.md
devshgraphicsprogramming: > Barriers that don't actually order anything can be elided. 

My issue isn't paying for barriers when there's nothing to be ordered.

Its more about accesses which look like they might be doing something that needs to be ordered but I didn't need them to be.

Sure the SPIR-V Aliasing rules really narrow down the set of accesses which could need ordering (except for BDA and SSBO, and workgroup if you explicitly lay it out), but those rules are extreme and there's quite a few problems with them as we've shown at SLS 2026.

> In the SPIR-V memory model, a "relevant access" would any "non-private" access; similarly in HLSL this would be any atomic or coherent access only.

hmmm, but in raw SPIR-V I can do OpLoad and OpStore with MakePointerAvailable and so on.

In fact I have to because VMM bans `Coherent` applied on a variable

On a SPIR-V level I'd expect those to be a "relevant access" if the storage class is non-private, especially that I'd avoid using OpAtomicLoad/OpAtomicStore when you don't have to worry about concurrent access (because there's none as clear happens-before relationships) and seeing partially updates values or having partial writes  as that would be like using a chainsaw for cutting grass.

Also there's a whole thing with OpAtomic only being usable on 4 byte types, and OpLoad/OpStore with MakeAvailable work on any type, and we suddenly get back to the nasty codegen for ByteAddressBuffers and friends that birthed the `untyped_pointer` extension to solve it

proposals/0007-mem-and-exec-model.md
devshgraphicsprogramming: > > There's probably a world in which you could imagine a front end compiler never omitting aliasing info; instead choosing to express the constraints solely with memory barriers. But we would need tighter barriers (on the order of affecting specific accesses) before that could be a practical reality.
> 
> ...I should probably raise an internal issue about this.

thats actually epic, should be written down as a perverse example/thought experiment in the docs. Really near and simple.

proposals/NNNN-mem-and-exec-model.md
Tobski: > Its more about accesses which look like they might be doing something that needs to be ordered but I didn't need them to be.

I think that warrants explicit tagging somehow. The default being safe I think is the better option.

> Sure the SPIR-V Aliasing rules really narrow down the set of accesses which could need ordering (except for BDA and SSBO, and workgroup if you explicitly lay it out), but those rules are extreme and there's quite a few problems with them as we've shown at SLS 2026.

SPIR-V aside, I think HLSL should define semantics for HLSL. Those should be well defined. Translating that to SPIR-V should be considered a separate problem.

> hmmm, but in raw SPIR-V I can do OpLoad and OpStore with MakePointerAvailable and so on.

> In fact I have to because VMM bans Coherent applied on a variable

I think we're digging into the weeds of SPIR-V too much for a HLSL proposal, but given you have to have NonPrivate alongside any AV/VIS flags I'm not sure what the problem is here?

proposals/0007-mem-and-exec-model.md
Tobski: There's some discussion happening now. I think what I want to get to is having aliasing layered on top of ordering in the memory model, and having the memory model have parity with that, but it'll take time. Anyway, again, off in the SPIR-V weeds when we're supposed to be talking about HLSL :)

proposals/NNNN-mem-and-exec-model.md
llvm-beanz: +💯 

> I think that warrants explicit tagging somehow. The default being safe I think is the better option.

Yes. This is one of our [codified design principles](https://github.com/hlsl-tc57/tc57/blob/main/docs/DesignConsiderations.md#portability-and-safety-by-default)!

> SPIR-V aside, I think HLSL should define semantics for HLSL. Those should be well defined. Translating that to SPIR-V should be considered a separate problem.

Yes. The goal in this spec is to define what the HLSL behavior should be. We have several members of TC57 who also participate in the Khronos SPIR and Vulkan working groups. We should define the language to enable user productivity and work through Khronos and the compiler implementers to ensure we can lower it to SPIRV (and DXIL).

proposals/NNNN-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+full portability and performance.
+
+## Motivation
+
+The HLSL and DXIL memory and execution model have never been fully defined. This
+forces reliance on Windows device certification for behavior conformance, and a
+general approach that the implementation defines the behavior.
+
+This state is not ideal from the start, however as HLSL becomes more widely
+portable, and DirectX moves to SPIRV the lack of a documented memory and
```
pow2clk: Comma not needed for a single and. Needed for the dependent clause.
```suggestion
portable and DirectX moves to SPIR-V, the lack of a documented memory and
```

proposals/NNNN-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+execution model makes it near impossible to ensure portability across byte code
+formats.
+
+The SPIRV-defined memory and execution model is constantly evolving and seeking
+to address some of the specific problems discussed in this proposal, however
+SPIRV is not designed to be written by humans. It is not a requirement that
+HLSL's memory and execution models match SPIRV, just that SPIRV is expressive
+enough to model programs in the model defined by HLSL.
```
pow2clk: Given that this is an HLSL spec, perhaps the onus should be placed on HLSL here. Where SPIR-V is unable to express what HLSL needs, changes will need to be made, but it might be better suited to this document to say something like 
```suggestion
SPIR-V is not designed to be written by humans. It is not a requirement that
HLSL's memory and execution models match SPIR-V, just that the HLSL model can be
expressed using SPIR-V.
```
It leaves open that SPIR-V might change, but also that HLSL needs to "fit" into it one way or another.

Also, I think we should spell it SPIR-V in formal documents

proposals/NNNN-mem-and-exec-model.md
pow2clk: That is what HLSL calls it

proposals/0007-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+  GroupMemoryBarrierWithGroupSync(); // sync point!
+  if (X == 0)
+    X = 2; // How many threads are in the tangle here?
+}
+```
+
+One problem with this model is not having clearly defined memory ordering which
+can impact tangle formation. If a thread is allowed to execute the second `if`
+body before all threads have finished evaluating the condition, tangle formation
+becomes unintuitive and potentially undefined.
```
pow2clk, line 136: Is this because the thread that got ahead might have rightly be grouped with some of the ones that haven't reached this point yet? That might bear mentioning. I can see how that would be unintuitive and potentially inefficient. I can't quite conceive of how it might be undefined unless you mean the tangle groupings would be effectively arbitrary

proposals/NNNN-mem-and-exec-model.md
pow2clk: Not sure why that would be. It seems atomic ops allow for exactly that kind of chaotic access

proposals/0007-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+(HSA)](https://en.wikipedia.org/wiki/Heterogeneous_System_Architecture)
+standards. This makes it reasonable that HLSL's memory model derive from HSA.
+
+Specific concerns that must be addressed:
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a wave?
+* What are the ordering requirements, if any, for memory operations to aliasing
+  memory across a set of tangled threads?
+
+## Appendix 1: Magic Decoder Ring
```
pow2clk: Colorful, but I'd prefer something clearer 
```suggestion
## Appendix 1: Glossary
```

proposals/0007-mem-and-exec-model.md
```diff
@@ -0,0 +1,162 @@
...
+
+| DirectX Term | Khronos Term | Description |
+| ------------ | ------------ | ----------- |
+| thread, lane | invocation   | The computation performed on a single element as described in the program. |
+|              | tangle       | A grouping of co-executing threads. |
```
pow2clk, line 160: I don't think this is specific enough. Arguably any group of threads running at the same time are co-executing
```suggestion
|              | tangle       | A grouping of threads co-executing the same instruction instance. |
```

proposals/NNNN-mem-and-exec-model.md
Tobski: @pow2clk you probably want to amend that last reference to "SPIRV"?

pow2clk: Yes, thanks!

proposals/0007-mem-and-exec-model.md
llvm-beanz: I think this is hard... getting more specific like your suggestion might be read that these threads are executing lock-step, which isn't a requirement. I'm also not entirely sure we should talk in terms of "instructions" since HLSL is really more about the "evaluation of expressions".

proposals/NNNN-mem-and-exec-model.md
llvm-beanz: > nit: s/warp/wave?

Fixed. HLSL does normatively use the term "wave" rather than "warp".

proposals/0007-mem-and-exec-model.md
llvm-beanz: It really depends on when and how the tangle formation is defined. Is it legal in the absence of explicit synchronization for an independent thread to proceed to the `if` statement, form a tangle of a single thread and continue executing?

If so, a second thread may reach the if statement _after_ the first thread sets X and not be part of the tangle.
If not, what rules exist that prevent it?

In particular do we need sync points for control flow statements? Or just control flow statements that access non-thread local memory?

Removing the `if` statement what about a case like:

```hlsl
groupshared int X;

[numthreads(4,1,1)]
void main(uint GI : SV_GroupIndex) {
  if (GI == 0)
    X = 0;
  GroupMemoryBarrierWithGroupSync(); // sync point!
  int MyVal = (X == 0) ? X = 2 : 0; // How many threads set X to 2?
}
```

Is this well-defined? What rules do we need to make in order to make it so?

proposals/0007-mem-and-exec-model.md
Tobski: What this model is going for as I've gathered from both reading and talking to @llvm-beanz is that you want the dynamic control flow to 1:1 match tangle formation, which we do not have in SPIR-V. Assuming that's achievable, you could perhaps say something more like "A grouping of threads executing a common control flow path" and work out from there. Somewhere along the way you need to talk about reconvergence, but that doesn't have to be detailed here.

bogner (approved):

bogner (approved):
I believe this is in good enough shape to go in so that we can start refining it. There is general agreement that we need to solve these problems, and the open questions and comments are about pinning down the specifics.

llvm-beanz:

PR #48, opened 2026-05-18
[New] Add numeric constants proposal

llvm-beanz:
This proposal suggests an alternative approach to HLSL's #INF syntax to avoid ambiguity in parsing against C preprocessor tokens.

### DIFF
--- a/proposals/0008-numeric-constants.md
+++ b/proposals/0008-numeric-constants.md
@@ -1,5 +1,5 @@
 ---
-title: NNNN - Numeric Constants
+title: 0008 - Numeric Constants
 params:
   authors:
   - llvm-beanz: Chris Bieneman
--- a/proposals/NNNN-numeric-constants.md
+++ b/proposals/NNNN-numeric-constants.md
@@ -1,0 +1,50 @@
+---
+title: NNNN - Numeric Constants
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+A new C++ STL-styled numeric constants interface will enable migrating off older
+inconsistent language features and make cleaner more concise code.
+
+## Motivation
+
+The HLSL Infinity constant syntax violates C token rules by using the `#`
+character, and is not consistent with C or C++ language styling, and is
+ambiguous against C preprocessor tokens. This causes technical challenges by
+using reserved preprocessor tokens, but also makes the feature foreign and less
+approachable in a C-like language.
+
+## Proposed solution
+
+Starting with adoption the `#INF` token becomes unsupported preferring instead a
+new library class modeled after the C++ `std::numeric_limits` classes. Below is
+the proposed interface for the `hlsl::numeric_limits` class:
+
+```c++
+  template<typename Ty>
+  class numeric_limits {
+  public:
+    static Ty min();
+    static Ty max();
+    static Ty lowest();
+    static Ty denorm_min();
+
+    // Implement infinity in terms of __builtin_huge_val
+    static Ty infinity();
+    static Ty negative_infinity();
+
+    static Ty quiet_NaN();
+    static Ty signaling_NaN();
+  }
+```
+
+This enables a straightforward code transformation from `1.#INF` to
+`hlsl::numeric_limits<float>::infinity()`. While the later is more verbose, it
+is more clear and provides a more consistent C++ style.


proposals/0008-numeric-constants.md
```diff
@@ -0,0 +1,50 @@
...
+    // Implement infinity in terms of __builtin_huge_val
+    static Ty infinity();
+    static Ty negative_infinity();
+
+    static Ty quiet_NaN();
+    static Ty signaling_NaN();
+  }
+```
+
+This enables a straightforward code transformation from `1.#INF` to
```
farzonl, line 48: What is our upgrade path for older code using `#INF`? Do we just let it error and have developers fix it or is this something we can teach clang-tidy to catch or clang-format to fix? Should we talk at all about the upgrade path in this proposal?

farzonl (approved):

pow2clk (approved):
Is this going to require constexpr as this comment suggests? https://github.com/microsoft/hlsl-specs/issues/103#issuecomment-2293187383

devshgraphicsprogramming:
> Is this going to require constexpr as this comment suggests? [microsoft/hlsl-specs#103 (comment)](https://github.com/microsoft/hlsl-specs/issues/103#issuecomment-2293187383)

One thing to note that `float` constexpr variables with exact bit-patterns one probably wants a `constexpr std::bit_cast` which requires c++20

llvm-beanz:
> Is this going to require constexpr as this comment suggests? [microsoft/hlsl-specs#103 (comment)](https://github.com/microsoft/hlsl-specs/issues/103#issuecomment-2293187383)

I don't believe this comment is accurate. I have separately proposed to add constexpr, but on-constexpr functions returning constant values will work just fine. Even libc++'s implementation of `std::numeric_limits<float>` can be compiled in C++ 03 mode without requiring `constexpr`.

> One thing to note that `float` constexpr variables with exact bit-patterns one probably wants a `constexpr std::bit_cast` which requires c++20

This is also not accurate on a couple levels. The libc++ implementation of `std::numeric_types<float>` does not rely on `bit_cast` it uses compiler builtins to provide bit-precise values for all the special cases. This pattern is trivial to follow in DXC.

Also, implementing constexpr bit_cast does not require C++20; we have it in Clang today under the HLSL mode which is not based on C++20. I also wouldn't use bit_cast for bit-precise values in this way except as a last resort. In my mind a better approach using C++17 would be to use hex floating point literals. They give exact patterns as literal values without needing casting. We could probably also support those in DXC, but none of that is required for the feature proposed here.

devshgraphicsprogramming:
> > Is this going to require constexpr as this comment suggests? [microsoft/hlsl-specs#103 (comment)](https://github.com/microsoft/hlsl-specs/issues/103#issuecomment-2293187383)
> 
> I don't believe this comment is accurate. I have separately proposed to add constexpr, but on-constexpr functions returning constant values will work just fine. Even libc++'s implementation of `std::numeric_limits<float>` can be compiled in C++ 03 mode without requiring `constexpr`.
> 
> > One thing to note that `float` constexpr variables with exact bit-patterns one probably wants a `constexpr std::bit_cast` which requires c++20
> 
> This is also not accurate on a couple levels. The libc++ implementation of `std::numeric_types<float>` does not rely on `bit_cast` it uses compiler builtins to provide bit-precise values for all the special cases. This pattern is trivial to follow in DXC.
> 
> Also, implementing constexpr bit_cast does not require C++20; we have it in Clang today under the HLSL mode which is not based on C++20. I also wouldn't use bit_cast for bit-precise values in this way except as a last resort. In my mind a better approach using C++17 would be to use hex floating point literals. They give exact patterns as literal values without needing casting. We could probably also support those in DXC, but none of that is required for the feature proposed here.

Clear Usecase / user story
```cpp
template<typename C, C val>
struct i_will_never_take_a_float_in_cpp20
{
    constexpr static inline C value = val;
};


int main()
{
    constexpr float f = i_will_never_take_a_float_in_cpp20<float,3.14159f>::value;
  std::cout << f << std::endl;

  return 0;
 }
```

hex literaly won't save me here, need C++20

can live without `std::bitcast` though

https://godbolt.org/z/cKeraqsce

llvm-beanz:
> Clear Usecase / user story
> 
> ```c++
> template<typename C, C val>
> struct i_will_never_take_a_float_in_cpp20
> {
>     constexpr static inline C value = val;
> };
> 
> 
> int main()
> {
>     constexpr float f = i_will_never_take_a_float_in_cpp20<float,3.14159f>::value;
>   std::cout << f << std::endl;
> 
>   return 0;
>  }
> ```
> 
> hex literaly won't save me here, need C++20
> 
> can live without `std::bitcast` though
> 
> https://godbolt.org/z/cKeraqsce

Okay... but this isn't a required case that this proposal is seeking to address, this is a different problem.

bogner (approved):

proposals/0008-numeric-constants.md
llvm-beanz: I think the actual upgrade path is a tooling problem out of the scope of the specification itself. It is certainly reasonable that DXC could provide fix-its or rewriter support.

llvm-beanz:

PR #51, opened 2026-05-18
[New] Add proposal to adopt modern C++

llvm-beanz:
This proposal adoptes C++11 as the base version of C++ for aligning the HLSL specification against. The proposal includes a curated set of C++11's features to include in the HLSL specification.


proposals/NNNN-cxx11-base.md
```diff
@@ -0,0 +1,48 @@
...
+* Lambda expressions
+* Static assert
+* Range-based for loops
+* Template parsing rules (no required space in `>>`)
+
+## Alternatives considered
+
+### C++20
+
+We could instead adopt an even more recent C++, like C++20. The main drawback of
```
farzonl: Why jump straight to 20 as the alt, why not consider 14 or 17 first? 

Second Is there something special about 11 or is it just the minimum feature set we want to support on 202y?

farzonl (approved):
LGTM

proposals/NNNN-cxx11-base.md
llvm-beanz: I've updated the framing of the proposal to instead iterate features and broaden out across C++14 and 17, because there may be good arguments to support some of those features as well.

proposals/0010-modern-cxx.md
```diff
@@ -0,0 +1,105 @@
+---
+title: NNNN - Modern C++ Features
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
```
pow2clk, line 8: In case you want to document the existing issues, which I think further motivate the proposal:

```suggestion

* Issues:
   * auto: [#24](https://github.com/microsoft/hlsl-specs/issues/24)
   * decltype: (#82)[https://github.com/microsoft/hlsl-specs/issues/82)
   * constexpr: [#21](https://github.com/microsoft/hlsl-specs/issues/21)
(#74)[https://github.com/microsoft/hlsl-specs/issues/74]
   * scoped enums: [#284](https://github.com/microsoft/hlsl-specs/issues/284)
   * variadic templates: [#21](https://github.com/microsoft/hlsl-specs/issues/21)
   * static assert: [#33](https://github.com/microsoft/hlsl-specs/issues/33)
   * simplified nested namespace: [#68](https://github.com/microsoft/hlsl-specs/issues/68)
   
```

Note that the constexpr issues aren't perfect matches. one is more specific and the other includes other suggestions. That's why 21 is repeated. 33 is ostensibly about a runtime assert, but most of the conversation is about static assert.

proposals/0010-modern-cxx.md
```diff
@@ -0,0 +1,105 @@
...
+* auto
+* decltype
+* constexpr
+* C++11 scoped enumerations
+* variadic templates
+* user-defined literals
+* Static assert
+* Template parsing rules (no required space in `>>`)
+
```
pow2clk, line 49: Is this the source?
```suggestion

[Source](https://en.cppreference.com/cpp/11)
```

proposals/0010-modern-cxx.md
```diff
@@ -0,0 +1,105 @@
...
+### C++14
+
+C++14 features that we could consider ([Source](https://en.cppreference.com/cpp/14)):
+
+* variable templates
+* relaxed restrictions on constexpr functions
+* binary literals
+* digit separators
+* return type deduction for functions
+
```
pow2clk, line 70:
```suggestion

[Source](https://en.cppreference.com/cpp/14)
```

pow2clk (approved):
Some suggested additions. I approve of all the content.

proposals/NNNN-cxx11-base.md
devshgraphicsprogramming: But C++20 has:
- concepts which are a lifesaver as demonstrated at SLS 2026
- designated initializers for structs
- consteval (I've been bitten by const static not really being compile time)
- std::bit_cast
- modules

IMHO modules would let us leapfrog PCH for reduced compilation times

proposals/0010-modern-cxx.md
llvm-beanz: I have the source here above, and I think that looks better in the rendering. I'll update them all that way.

proposals/NNNN-cxx11-base.md
llvm-beanz: From a practical perspective, DXC is never going to support any features from C++20 that require more than a few hours of engineering to build and test. That rules out everything you've listed here except maybe `bit_cast`.

TC57's decision then is do we want to include language in the specification that the only currently maintained production HLSL compiler can never support? My gut feeling is no, but we can discuss at the next meeting.

I think modules in particular is an interesting example of the problem here. Even today Clang does not fully support C++20 modules. In fact, the only compiler that claims full support for Modules is MSVC. Modules adoption has been massively hampered by the fact that the feature has minimal support in most toolchains (https://arewemodulesyet.org).

Were TC57 to put modules in the HLSL specification there would be no implementation that supported it until at the earliest Clang's HLSL implementation is feature-complete. This assumes that Clang's implementation of C++ modules is complete by then. Even given that, there are almost certainly other changes that would need to be made in order for C++ modules to be compatible with HLSL, and that work would go beyond what is already scoped for the HLSL implementation in Clang. I would be hesitant to say we could have an HLSL implementation with modules before the end of the decade.

I would not recommend that TC57 include such a large feature in the initial draft of the HLSL specification since it would considerably delay both the completion of the specification and the existence of any conforming implementations.

devshgraphicsprogramming: okay that clears up stuff for me a lot, the first TC57 formalizes what DXC will support not Clang-HLSL

llvm-beanz: Clang will also implement the standard from the initial version, and subsequent revisions.

proposals/NNNN-modern-cxx.md
```diff
@@ -0,0 +1,113 @@
...
+* new order of evaluation rules (these changes don't apply to HLSL)
+* lambda capture of *this (requires lambdas)
+* constexpr lambda (requires lambdas)
+* inline variables (requires linkage model changes)
+* guaranteed copy elision (requires construction & rvalue references)
+
+### C++20 and later
+
+Adopting C++ 20 and later features is explicitly rejected due to feasibility in
+existing implementations.
```
bogner: This wording sounds like we're never going to consider taking c++20 or later features at all, which isn't the intent. Maybe:
```suggestion
Adopting C++ 20 and later features is not under consideration at this time due
to feasibility in existing implementations.
```

bogner (approved):

llvm-beanz:

PR #52, opened 2026-05-18
[0002] Add specification language for literals

### LINKED ISSUE #31: [0002] Consider suffixes for 16bit types
**Which proposal does this relate to?**
[0002 - Conforming Literals](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0002-conforming-literals.md)

**Describe the issue or outstanding question.**
Consider adding suffixes for 16-bit types.

**Additional context**

https://github.com/hlsl-tc57/tc57/pull/15#issuecomment-4250137738

llvm-beanz:
This updates the conforming literals proposal with detailed specification language.

Fixes #31


proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,126 @@ users to assist in identifying behavior changes.
...
 ## Detailed Design
 
-The full proposed specification for floating point literals in HLSL is in
-[#175](https://github.com/microsoft/hlsl-specs/pull/175).
-
-The full proposed specification for integer literals in HLSL is in
-[#208](https://github.com/microsoft/hlsl-specs/pull/208).
+### Literals [Lex.Literal]
+
+```latex
```
farzonl, line 264: this latex block does not actually cause this to show up as LaTex in the Rich diff display. Could we add a picture for this showing what the new spec language would look like? It's a little hard to follow pure LaTex.

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,126 @@ users to assist in identifying behavior changes.
...
+  representation size.
+*  The rank of a minimum precision integer type shall be less than the rank
+  of all standard integer types.
+*  For all integer types `T1`, `T2`, and `T3`: if
+  `T1` has greater rank than `T2` and `T2` has greater rank
+  than `T3`, then `T1` shall have greater rank than `T3`.
+
+#### Floating Point Conversion Rank [Conv.rank.float]
+
+*  The rank `half` shall be greater than the rank of `min16float`.
```
pow2clk, line 422: well I guess I have to duplicate my concern expressed in https://github.com/hlsl-tc57/tc57/pull/24#discussion_r3278464265 to here

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,126 @@ users to assist in identifying behavior changes.
...
+  vector-literal\br
+
+  \define{integer-literal}\br
+  decimal-literal \opt{integer-suffix}\br
+  octal-literal \opt{integer-suffix}\br
+  hexadecimal-literal \opt{integer-suffix}\br
+
+  \define{decimal-literal}\br
+  nonzero-digit\br
+  decimal-literal digit\br
```
pow2clk: Huh. Does this mean all `0` are technically octal literals? I think that's fine, just never thought of it.

proposals/0002-conforming-literals.md
pow2clk: If it would help you to review it in the meantime, there is only one difference from this https://timsong-cpp.github.io/cppwp/n3337/lex.literal

Instead of user-defined-literals, we have vector-literals.

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,126 @@ users to assist in identifying behavior changes.
...
+  unsigned-suffix \opt{long-suffix}\br
+  long-suffix \opt{unsigned-suffix}\br
+
+  \define{unsigned-suffix} \textnormal{one of}\br
+  \terminal{u U}\br
+
+  \define{long-suffix} \textnormal{one of}\br
+  \terminal{l L}
+\end{grammar}
+```
```
pow2clk, line 321: I think this was meant to be in the Integer Literals section, but it's just above the start of that section.

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,126 @@ users to assist in identifying behavior changes.
...
+
+An implementation may support the integer suffixes `ll` and
+`ull` as equivalent to `l` and `ul` respectively.
+
+### Floating Point Literals [Lex.Literal.Float]
+
+A floating literal is written either as a _fractional-constant_ with
+an optional _exponent-part_ and optional _floating-suffix_, or as
+an integer _digit-sequence_ with a required _exponent-part_ and
+optional _floating-suffix_.
```
pow2clk, line 383: Did you want to include the grammar for floats here as well? Otherwise these terms aren't defined.

pow2clk (commented):
Duplicate 16-bit concern from the Conversions clause and a couple grammar section issues

proposals/0002-conforming-literals.md
llvm-beanz: I tried embedded css and html... didn't work, so I've included the images.

proposals/0002-conforming-literals.md
llvm-beanz: Yep. That's a fun oddity of the C grammar.

proposals/0002-conforming-literals.md
llvm-beanz: I added a note here that the conversion ranks aren't really _changed_ by this proposal. They just apply to literals which they didn't in HLSL 2021 because of the `literal` types. We should use the conversions PR to address any concerns with this language.

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +254,162 @@ users to assist in identifying behavior changes.
...
+  \texttt{E} \opt{sign} digit-sequence\br
+  \define{sign} \textnormal{one of}\br
+  \texttt{+} \texttt{-}\br
+  \define{digit-sequence}\br
+  digit\br
+  digit-sequence digit
+  \define{floating-suffix} \textnormal{one of}
+  \texttt{h} \texttt{f} \texttt{l} \texttt{H} \texttt{F} \texttt{L}
+\end{grammar}
+```
```
devshgraphicsprogramming, line 376: what are the reasons against supporting the C++23 floating point suffices?

llvm-beanz: We have an issue tracking that (https://github.com/hlsl-tc57/tc57/issues/31), but haven't discussed it in the committee yet.

llvm-beanz: In general, we've avoided adding C++17 and later features because DXC's fork of Clang has only partial C++17 support and no support for C++20+, so any feature we pull into HLSL from later versions of C++ requires full implementation in DXC.

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +259,168 @@ users to assist in identifying behavior changes.
...
+  \terminal{u U}\br
+
+  \define{long-suffix} \textnormal{one of}\br
+  \terminal{l L}
+\end{grammar}
+```
+
+![integer-literal](0002-assets/integer-literal.png)
+
+An _integer-literal_ is an optional base prefix, a sequence of digits
```
V-FEXrt: It seems like the `base prefix` is missing from the grammar and only mentioned here? Do we need to add language about it?

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +259,168 @@ users to assist in identifying behavior changes.
...
+
+The type of an integer literal is the first of the corresponding list in the
+table below in which its value can be represented.
+
+| Suffix                         | Decimal constant       | Octal or hexadecimal constant                |
+|--------------------------------|------------------------|----------------------------------------------|
+| none                           | `int32_t`, `int64_t`   | `int32_t`, `uint32_t`, `int64_t`, `uint64_t` |
+| `u` or `U`                     | `uint32_t`, `uint64_t` | `uint32_t`, `uint64_t`                       |
+| `l` or `L`                     | `int64_t`              | `int64_t`, `uint64_t`                        |
+| Both `u` or `U` and `l` or `L` | `uint64_t`             | `uint64_t`                                   |
```
V-FEXrt, line 337: Are `uL` and `Ul` intentionally allowed here?

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +259,168 @@ users to assist in identifying behavior changes.
...
+  octal-literal \opt{integer-suffix}\br
+  hexadecimal-literal \opt{integer-suffix}\br
+
+  \define{decimal-literal}\br
+  nonzero-digit\br
+  decimal-literal digit\br
+
+  \define{octal-literal}
+  \terminal{0}\br
+  octal-literal octal-digit\br
```
V-FEXrt: I think this is missing the octal-literal terminal prefix?

V-FEXrt (approved):
Approving but you should make sure other reviewers feel their comments are addressed!

proposals/0002-conforming-literals.md
llvm-beanz: This is confusing, but correct via recursion. Because an _octal-literal_ must be either `0` or "_octal-literal_ _octal-digit_" the only way the recursive expansion ends is if the leading digit is a `0`, which is the octal prefix.

proposals/0002-conforming-literals.md
llvm-beanz: Yep, C (and HLSL today) allow mixing case (https://godbolt.org/z/vKz7zxKas)

proposals/0002-conforming-literals.md
```diff
@@ -254,15 +259,168 @@ users to assist in identifying behavior changes.
...
 
-The full proposed specification for integer literals in HLSL is in
-[#208](https://github.com/microsoft/hlsl-specs/pull/208).
+  \define{hexadecimal-literal}\br
+  \terminal{0x} hexadecimal-digit\br
+  \terminal{0X} hexadecimal-digit\br
+  hexadecimal-literal hexadecimal-digit\br
+
+  \define{nonzero-digit} \textnormal{one of}\br
+  \terminal{1 2 3 4 5 6 7 8 9}\br
```
bob80905: Strange, the 1 is not showing up in the png.

llvm-beanz: Yea... I just posted a PR to fix the HTML rendering:
https://github.com/hlsl-tc57/tc57/pull/59

proposals/0002-conforming-literals.md
llvm-beanz: I've updated the language around this to match the grammar formulations.

proposals/0002-conforming-literals.md
llvm-beanz: I've updated the images so the pngs should be correct now.

V-FEXrt (approved):

PR #53, opened 2026-05-22
[Infra] Fix docker target names

llvm-beanz:
I somehow ended up committing these in a weird broken state. This just fixes the target names so that it all works correctly.


Alexander-Johnston (approved):
Tested locally, fix seems good.

PR #54, opened 2026-05-25
[New] Add proposal for strict initializer lists

llvm-beanz:
This proposal adopts conformant initializer list behavior to C++. This proposal is required to properly make use of user-defined conversion functions, constructors, and other frequently requested behaviors.

### DIFF
--- a/proposals/0011-strict-initializer-lists.md
+++ b/proposals/0011-strict-initializer-lists.md
@@ -1,5 +1,5 @@
 ---
-title: NNNN - Strict Initializer Lists
+title: 0011 - Strict Initializer Lists
 params:
   authors:
   - llvm-beanz: Chris Bieneman
@@ -9,7 +9,8 @@ params:
 ---
 
 * Planned Version: 202y
-* Dependencies: [0003 Overload Resolution](0003-overload-resolution.md
+* Dependencies: [0003 Overload Resolution](0003-overload-resolution.md)
+
 ## Introduction
 
 HLSL should adopt C and C++ initializer list parsing rules, rather than
--- a/proposals/NNNN-strict-initializer-lists.md
+++ b/proposals/NNNN-strict-initializer-lists.md
@@ -10,7 +10,7 @@ params:
 
 ## Introduction
 
-HLSL should adopt C and C++ initializer list formatting rules, rather than
+HLSL should adopt C and C++ initializer list parsing rules, rather than
 custom rules that are unintuitive and error prone.
 
 ## Motivation
--- a/proposals/NNNN-strict-initializer-lists.md
+++ b/proposals/NNNN-strict-initializer-lists.md
@@ -8,6 +8,8 @@ params:
   status: Under Consideration
 ---
 
+* Planned Version: 202y
+* Dependencies: [0003 Overload Resolution](0003-overload-resolution.md
 ## Introduction
 
 HLSL should adopt C and C++ initializer list parsing rules, rather than
--- a/proposals/NNNN-strict-initializer-lists.md
+++ b/proposals/NNNN-strict-initializer-lists.md
@@ -1,0 +1,116 @@
+---
+title: NNNN - Strict Initializer Lists
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+HLSL should adopt C and C++ initializer list formatting rules, rather than
+custom rules that are unintuitive and error prone.
+
+## Motivation
+
+HLSL supports flattened brace initialization such that the structure of the
+bracket initializer on the right hand side of an initialization is ignored, and
+only the number of initialization arguments matters. Further, vector and matrix
+arguments are implicitly expanded.
+
+This feature likely results in a variety of common errors as HLSL will attempt
+to fit an initializer list to a structure regardless of the underlying structure
+of the members.
+
+In HLSL the following code is valid:
+
+```c++
+  struct A {
+    int a;
+    double b;
+  };
+  struct B {
+    A a[2];
+    int c;
+  };
+  B b = {{1, 1.2}, {2, 2.2}, 3};   // Array elements specified as members
+  B b2 = {1, 2, 3, 4, 5};          // each field initialized separately
+  B b3 = {{1, {2, 3}}, {4, 5}};    // Completely random grouping of arguments
+  int4 i4 = {1,2,3,4};             // valid int4 in C-syntax
+  B b4 = {i4, 5};                  // int4 implicitly expanded to 4 arguments
+```
+
+Formalizing this code to comply with C/C++ initializer list rules will likely
+break existing code, however following C/C++ rules will allow error checking and
+validation of initializer lists which is likely to catch bugs which may be
+difficult to identify otherwise.
+
+## Proposed solution
+
+Adopt C & C++ initializer list rules, and remove HLSL-specific initialization
+list behaviors. Specifically, this will remove implicit vector and structure
+element extraction, and enforce type system rules for data types passed into
+initializer lists.
+
+This will also introduce C rules for zero-initialization including zero
+initialization for structure members omitted from the initialization list.
+
+### Forward-source breaking
+
+This change will be forward source breaking, and backwards compatible with some
+caveats. Current code that takes advantage of HLSL initialization semantics will
+produce an error, but implementations should be able to automate migration
+throuhg tooling. Code updated to support the new syntax should be mostly
+backward compatible to older HLSL versions as long as each structure is fully
+initialized.
+
+For example given the source example from the motivation section above, a tool
+could rewrite the initializers as:
+
+```c++
+  B b = {{{1, 1.2}, {2, 2.2}}, 3};
+  B b2 = {{{1, 2}, {3, 4}}, 5};
+  B b3 = {{{1, 2}, {3, 4}}, 5};
+  int4 i4 = {1,2,3,4};
+  B b4 = {{{i4.x, i4.y}, {i4.z, i4.w}}, 5};
+```
+
+The rewritten initializers all behave the same, but shift braces to match
+structure layouts and explicitly expand aggregate members.
+
+### Non-backward compatible uses
+
+New code that takes advantage of C & C++'s zero-initialization behavior or
+relies on user-defined convesrion functions will not be backwards compatible to
+older HLSL versions.
+
+For example:
+
+```c++
+struct S {
+  int X[2];
+  operator float() {
+    return ((float)X[0])/((float)X[1]));
+  }
+
+  operator double() {
+    return (float)this;
+  }
+};
+
+struct F {
+  float2 f;
+};
+
+struct D {
+  double3 f;
+};
+
+S s1 = {{1,2}}; // Valid with and without this change.
+S s2 = {};      // Invalid without and valid with this proposal, zero-initializes all members.
+S s3 = {1};     // Invalid without and valid with this proposal, zero-initializes unspecified members.
+F f1 = {s1};    // Invalid without and valid in both cases, but the meaning changes. Without this proposal f1 is initialized to {1,2}, with this proposal it is initialized to {0.5, 0}.
+D d1 = {s1};    // Invalid without and valid with this proposal, initializes to {0.5,0,0}.
+```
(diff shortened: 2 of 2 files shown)


proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+HLSL should adopt C and C++ initializer list formatting rules, rather than
+custom rules that are unintuitive and error prone.
```
bogner: I don't think formatting is the right word here.
```suggestion
HLSL should adopt C and C++ initializer list parsing rules, rather than
custom rules that are unintuitive and error prone.
```

bogner (approved):

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+HLSL should adopt C and C++ initializer list parsing rules, rather than
+custom rules that are unintuitive and error prone.
```
pow2clk: Nit. The internet seems to agree this takes a hyphen
```suggestion
custom rules that are unintuitive and error-prone.
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Motivation
+
+HLSL supports flattened brace initialization such that the structure of the
+bracket initializer on the right hand side of an initialization is ignored, and
```
pow2clk:
```suggestion
bracket initializer on the right-hand side of an initialization is ignored, and
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+only the number of initialization arguments matters. Further, vector and matrix
+arguments are implicitly expanded.
+
+This feature likely results in a variety of common errors as HLSL will attempt
+to fit an initializer list to a structure regardless of the underlying structure
+of the members.
+
+In HLSL the following code is valid:
+
+```c++
```
pow2clk: hlsl is a valid language selection for md in github https://github.com/github-linguist/linguist/blob/e9fe3c9f230cd9220afcd057f75702de4d7700c9/lib/linguist/languages.yml#L3002
```suggestion
```hlsl
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+caveats. Current code that takes advantage of HLSL initialization semantics will
+produce an error, but implementations should be able to automate migration
+throuhg tooling. Code updated to support the new syntax should be mostly
+backward compatible to older HLSL versions as long as each structure is fully
+initialized.
+
+For example given the source example from the motivation section above, a tool
+could rewrite the initializers as:
+
+```c++
```
pow2clk:
```suggestion
```hlsl
```

proposals/0011-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+break existing code, however following C/C++ rules will allow error checking and
+validation of initializer lists which is likely to catch bugs which may be
+difficult to identify otherwise.
+
+## Proposed solution
+
+Adopt C & C++ initializer list rules, and remove HLSL-specific initialization
+list behaviors. Specifically, this will remove implicit vector and structure
+element extraction, and enforce type system rules for data types passed into
+initializer lists.
```
pow2clk, line 57: This means that this common usage will be disallowed:

```hlsl
float4 main (float3 LightDir : LVec) : SV_Target {
  float4 HomLightDir = float4(LightDir, 1.0);
...
```
With this change. I can see no way to do this other than:
```hlsl
float4 main (float3 LightDir : LVec) : SV_Target {
  float4 HomLightDir = float4(LightDir.x, LightDir.y, LightDir.z, 1.0);
...
```
Which is a lot more clunky.

Perhaps something where if the vector matches the next target element type perfectly, than it gets copied directly over, otherwise, it extracts the types as the vector's elements and applies them to the current initialization scope

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+
+This will also introduce C rules for zero-initialization including zero
+initialization for structure members omitted from the initialization list.
+
+### Forward-source breaking
+
+This change will be forward source breaking, and backwards compatible with some
+caveats. Current code that takes advantage of HLSL initialization semantics will
+produce an error, but implementations should be able to automate migration
+throuhg tooling. Code updated to support the new syntax should be mostly
```
pow2clk: typo
```suggestion
through tooling. Code updated to support the new syntax should be mostly
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+
+```c++
+  B b = {{{1, 1.2}, {2, 2.2}}, 3};
+  B b2 = {{{1, 2}, {3, 4}}, 5};
+  B b3 = {{{1, 2}, {3, 4}}, 5};
+  int4 i4 = {1,2,3,4};
+  B b4 = {{{i4.x, i4.y}, {i4.z, i4.w}}, 5};
+```
+
+The rewritten initializers all behave the same, but shift braces to match
```
pow2clk:
```suggestion
The rewritten initializers all behave the same but shift braces to match
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+structure layouts and explicitly expand aggregate members.
+
+### Non-backward compatible uses
+
+New code that takes advantage of C & C++'s zero-initialization behavior or
+relies on user-defined convesrion functions will not be backwards compatible to
```
pow2clk: typo
```suggestion
relies on user-defined conversion functions will not be backwards compatible to
```

pow2clk: User-defined conversion operator overloads are not part of this proposal

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+
+struct D {
+  double3 f;
+};
+
+S s1 = {{1,2}}; // Valid with and without this change.
+S s2 = {};      // Invalid without and valid with this proposal, zero-initializes all members.
+S s3 = {1};     // Invalid without and valid with this proposal, zero-initializes unspecified members.
+F f1 = {s1};    // Invalid without and valid in both cases, but the meaning changes. Without this proposal f1 is initialized to {1,2}, with this proposal it is initialized to {0.5, 0}.
+D d1 = {s1};    // Invalid without and valid with this proposal, initializes to {0.5,0,0}.
```
pow2clk: First, this would require conversion operator overloading to work, which this doesn't specify. Second, it brings up a dicey question. In this case, would we splat the scalar to the vector it is initializing or would it do the zero initialization? I think it may actually be missing a set of braces as it is using an S struct presumably cast to a double to initialize a 3-element vector of doubles, which would require another layer of braces. In that case, I don't see any ambiguity as assigning a scalar to a vector is established as being a splat and assigning only one element of the innermost {} representing the vector will have implicit initialization to zero with this proposal
```suggestion
D d1 = {{s1}};    // Invalid without and valid with this proposal, initializes to {0.5,0,0}.
```

proposals/NNNN-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+relies on user-defined convesrion functions will not be backwards compatible to
+older HLSL versions.
+
+For example:
+
+```c++
+struct S {
+  int X[2];
+  operator float() {
+    return ((float)X[0])/((float)X[1]));
```
pow2clk: Extra paren
```suggestion
    return ((float)X[0])/((float)X[1]);
```

proposals/0011-strict-initializer-lists.md
```diff
@@ -0,0 +1,116 @@
...
+
+```c++
+struct S {
+  int X[2];
+  operator float() {
+    return ((float)X[0])/((float)X[1]));
+  }
+
+  operator double() {
+    return (float)this;
```
pow2clk, line 100: A reminder that `this` is a reference in HLSL might make this look less very wrong to a C++ programmer

proposals/NNNN-strict-initializer-lists.md
llvm-beanz: I think in this case it's fine, but I find the `hlsl` markdown formatting is fine for 2018, but not 2021+.

proposals/0011-strict-initializer-lists.md
llvm-beanz: This is really interesting point. I hadn't thought about it, but zero-extending and one-extending vectors is pretty common. We probably should think about some short-hands for that.

proposals/NNNN-strict-initializer-lists.md
llvm-beanz: They are in [0003 - overload resolution](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0003-overload-resolution.md), which is also under consideration.

proposals/NNNN-strict-initializer-lists.md
llvm-beanz: Conversion operators are in proposal [0003 - overload resolution](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0003-overload-resolution.md).

C/C++ rules don't strictly require nested braces, and I don't think braces would change the meaning of the expression in any case because providing a single scalar as the inner initializer with and without braces behaves the same.

This does raise a question though about whether we this is initializing the first element of the vector through a user-defined conversion then zero-initializing the rest of the structure (as I initially wrote it thinking) or if the secondary standard-conversion sequence after the user-defined conversion performs a splat. I _think_ the later is more correct having just written the language for this in the overload resolution of user-defined conversions. I'll update.

proposals/0011-strict-initializer-lists.md
llvm-beanz: Apologies for not addressing this point explicitly. I'm curious for feedback from the wider committee on this. I worry that if we remind people of things like this every time we write about them in our spec and proposal documents we're going to end up with documents that just constantly repeat over and over again "this is different from C++", which I think detracts from the utility of the specification.

proposals/NNNN-strict-initializer-lists.md
pow2clk: I think including user-defined conversion operators as part of a proposal whose main intent is to change the way existing overloads behave is packing too many things into an extension, but either way, no dependency is not represented in this document

proposals/NNNN-strict-initializer-lists.md
pow2clk: Whether this is splatted or zero-initialized is exactly the question I was trying to raise. It seems to me that `{s1}` should be splatted because it's providing a scalar to initialize a vector and that has always resulted in a splat and `{{s1}}` should be zero-initialized since it is providing one scalar where a full initialization would include three (e.g.`{{s1, s1, 1.0}}`)

proposals/0011-strict-initializer-lists.md
jessebarker: Could we have a section analogous to "errata" that calls out specific differences?  Or will they really be too numerous to maintain cleanly?

jessebarker (approved):

pow2clk:
I want to see the lingering issues with initialization of vectors addressed before this is accepted. I'm approving this going in as-is with the understanding that they will be addressed

pow2clk (approved):

pow2clk:
I'd also like to see the dependencies and targeted version added

llvm-beanz:

proposals/0011-strict-initializer-lists.md
```diff
@@ -0,0 +1,135 @@
+---
+title: NNNN - Strict Initializer Lists
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
```
pow2clk, line 10:
```suggestion

* Planned Version: 202y
* Dependencies: [0003 Overload Resolution](0003-overload-resolution.md
```

PR #57, opened 2026-05-27
Implementation status for in-progress proposals

llvm-beanz:
This PR adds implementation status for proposals that are in progress by engineering teams.

### DIFF
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -8,6 +8,13 @@ params:
   status: Refinement
 ---
 
+## Implementation Status
+
+|   | DXC     | Clang    |
+|---|---------|----------|
+| C++-aligned best match | Not Planned | Complete |
+| C++-aligned candidate selection | Not Planned | Complete |
+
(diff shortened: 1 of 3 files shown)


Alexander-Johnston:
Should we not have a status for https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md too?

llvm-beanz:
> Should we not have a status for https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md too?

I left that off because I don't believe we've started that work in any compiler yet.

Alexander-Johnston:
> > Should we not have a status for https://github.com/hlsl-tc57/tc57/blob/main/proposals/0006-restricted-unbound-arrays.md too?
> 
> I left that off because I don't believe we've started that work in any compiler yet.

Makes sense. I suppose if people do ask about it in the future we can add a "Planned but unstarted" or something to that end.

Alexander-Johnston (approved):

PR #59, opened 2026-05-29
Fix pandoc grammar rendering

llvm-beanz:
The grammar rendering by pandoc was broken in the change to add Annex A because pandoc doesn't read macros out of nested environments correctly. This fixes the issue by declaring the macros golbally. It's a bit sad, but fine.


Alexander-Johnston (approved):
Works for me now.

PR #60, opened 2026-05-29
Add a narrow-display mode for the side bar

llvm-beanz:
This dramatically improves rendering on mobile devices and allows for rendering in narrow windows on other devices as well.

Assisted by copilot.


Alexander-Johnston (approved):
Playing with this, it's so much nicer on mobile.

I went to see how a few other specs render on mobile too and noticed in the Vulkan spec that on a narrow screen it still introduces a bunch of horizontal scroll, and in turn lots of dead space, to accommodate wide tables.
This doesn't seem to be a problem for us yet - notes, grammars, and the like all seem to rearrange themselves and flow nicely for narrow screens too, but I think we should keep it in the back of our heads. We might need to implement some more css for tables to give them a max width and their own scroll bars on mobile.

PR #62, opened 2026-06-01
[0003] Add detailed spec language for Overloading

### LINKED ISSUE #92: [0003] Detailed Design
This tracks completing and merging the detailed design for [[0003] Overload Resolution](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0003-overload-resolution.md)

llvm-beanz:
This fleshes out the contents of what will become the Overloading clause.

Resolves #92

### DIFF
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -512,9 +512,7 @@ below:
 | Array-to-pointer | Lvalue Transformation | Exact Match | \ref{Conv.array} |
 | Qualification | Qualification Adjustment | Exact Match | \ref{Conv.qual} |
 | Vector Scalar splat (without conversion) | Scalar Extension | Extension | \ref{Conv.vsplat} |
-| Vector Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
 | Matrix Scalar splat (without conversion) | Scalar Extension | Extension | \ref{Conv.msplat} |
-| Matrix Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.msplat} |
 | Integral promotion | Promotion | Promotion | \ref{Conv.iconv} & \ref{Conv.rank.int} |
 | Floating point promotion | Promotion | Promotion | \ref{Conv.fconv} & \ref{Conv.rank.float} |
 | Component-wise promotion | Promotion | Promotion | \ref{Conv.cwise} |
@@ -526,11 +524,13 @@ below:
 | Boolean conversion | Conversion | Conversion | \ref{Conv.bool} |
 | Component-wise conversion | Conversion | Conversion | \ref{Conv.cwise} |
 | Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
+| Vector Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
+| Matrix Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.msplat} |
 | Vector truncation (without conversion) | Dimensionality Reduction | Truncation | \ref{Conv.vtrunc} |
-| Vector truncation promotion | Dimensionality Reduction Promotion | Promotion Truncation | \ref{Conv.vtrunc} |
-| Vector truncation conversion | Dimensionality Reduction Conversion | Conversion Truncation | \ref{Conv.vtrunc} |
 | Matrix truncation (without conversion) | Dimensionality Reduction | Truncation | \ref{Conv.vtrunc} |
+| Vector truncation promotion | Dimensionality Reduction Promotion | Promotion Truncation | \ref{Conv.vtrunc} |
 | Matrix truncation promotion | Dimensionality Reduction Promotion | Promotion Truncation | \ref{Conv.vtrunc} |
+| Vector truncation conversion | Dimensionality Reduction Conversion | Conversion Truncation | \ref{Conv.vtrunc} |
 | Matrix truncation conversion | Dimensionality Reduction Conversion | Conversion Truncation | \ref{Conv.vtrunc} |
 
 If a scalar splat conversion occurs in a conversion sequence where all other
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -678,12 +678,24 @@ subscript syntax when the _postfix-expression_ is an object of class type.
 
 #### Built-in Operators `[Overload.Builtin]`
 
-// arithmetic
+For binary operators on arithmetic types, the result type is called the
+_promoted expression type_ as determined by usual arithmetic conversions
+(\ref{Expr.Conv}).
+
+For all types `T` where `T` is an arithmetic type (\ref{Basic.Types.Arithmetic});
+operators will be defined of the forms:
+
+```
 T operator+(T );
 T operator-(T );
 T operator~(T );
+```
 
-// arithmetic
+For all types `L`, `R` and `LR`, where `L` and `R` are any arithmetic type
+(\ref{Basic.Types.Arithmetic}), and `LR` is the promoted expression type;
+operators will be defined of the forms:
+
+```
 LR operator*(L , R );
 LR operator/(L , R );
 LR operator+(L , R );
@@ -695,42 +707,73 @@ bool operator>=(L , R );
 bool operator==(L , R );
 bool operator!=(L , R );
 LR operator%(L , R );
+```
 
-// enumeration
+For all types `T` where `T` is an enumeration type (\ref{Decl.Enum}); operators
+will be defined of the form:
+
+```
 bool operator<(T , T );
 bool operator>(T , T );
 bool operator<=(T , T );
 bool operator>=(T , T );
 bool operator==(T , T );
 bool operator!=(T , T );
+```
 
-// integral
+For all types `T` where `T` is an integral type
+(\ref{Basic.Types.Arithmetic}); operators will be defined of the form:
+
+```
 LR operator&(L , R );
 LR operator^(L , R );
 LR operator|(L , R );
 L operator<<(L , R );
 L operator>>(L , R );
+```
 
-// arithmetic
-VQ L & operator=(VQ L &, R );
-VQ L & operator*=(VQ L &, R );
-VQ L & operator/=(VQ L &, R );
-VQ L & operator+=(VQ L &, R );
-VQ L & operator-=(VQ L &, R );
-VQ L & operator%=(VQ L &, R );
+For all types `L` and `R` where `L` and `R` are arithmetic types
+(\ref{Basic.Types.Arithmetic}), and `CVQ` is a _cv-qualifier_; operators will be
+defined of the form:
 
-// integral
-VQ L & operator<<=(VQ L &, R );
-VQ L & operator>>=(VQ L &, R );
-VQ L & operator&=(VQ L &, R );
-VQ L & operator^=(VQ L &, R );
-VQ L & operator|=(VQ L &, R );
+```
+CVQ L & operator=(CVQ L &, R );
+CVQ L & operator*=(CVQ L &, R );
+CVQ L & operator/=(CVQ L &, R );
+CVQ L & operator+=(CVQ L &, R );
+CVQ L & operator-=(CVQ L &, R );
+CVQ L & operator%=(CVQ L &, R );
+```
 
+For all types `L` and `R` where `L` and `R` are integral types
+(\ref{Basic.Types.Arithmetic}), and `CVQ` is a _cv-qualifier_; operators will be
+defined of the form:
+```
+CVQ L & operator<<=(CVQ L &, R );
+CVQ L & operator>>=(CVQ L &, R );
+CVQ L & operator&=(CVQ L &, R );
+CVQ L & operator^=(CVQ L &, R );
+CVQ L & operator|=(CVQ L &, R );
+```
+
+For the boolean type `bool` operators will be defined of the form:
+
+```
 bool operator!(bool);
 bool operator&&(bool, bool);
 bool operator||(bool, bool);
+```
 
-// arithmetic
+For all types `L`, `R` and `LR`, where `L` and `R` are any arithmetic type
+(\ref{Basic.Types.Arithmetic}), and `LR` is the promoted expression type;
+the ternary operator will be defined of the form:
+
+```
 LR operator?:(bool, L , R );
+```
 
+For all types `T`, a ternary operator will be defined of the form:
+
+```
 operator?:(bool, T , T );
+```
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -9,7 +9,8 @@ params:
 ---
 
 * Dependencies: [#74](https://github.com/hlsl-tc57/tc57/pull/74),
-                [#115](https://github.com/hlsl-tc57/tc57/pull/115)
+                [#115](https://github.com/hlsl-tc57/tc57/pull/115),
+                [#127](https://github.com/hlsl-tc57/tc57/pull/127)
 
 ## Implementation Status
 
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -159,15 +159,13 @@ uint Yeet(); // ill-formed: decls differ only by return type
 ```hlsl
 class Doggo {
   static void pet();
-  void pet();              // ill-formed: static pet has the same parameter-type-list
-  void pet() const;        // ill-formed: static pet has the same parameter-type-list
+  void pet();              // ill-formed: cannot be overloaded with `static void pet()`
 
-  void wagTail();          // valid: no conflicting static declaration.
-  void wagTail() const;    // valid: no conflicting static declaration.
+  static void wagTail();
+  void wagTail();          // valid
 
-  static void bark(Doggo D);
-  void bark();             // valid: static bark parameter-type-list is different
-  void bark() const;       // valid: static bark parameter-type-list is different
+  void bark(int);
+  static void bark(int);   // ill-formed: cannot be overloaded with `void bark(int)`
 };
 ```
 
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -622,7 +622,7 @@ rules applies:
 operator-function-id:
   `operator` operator
 operator: one of
-  `+` `-` `*` `/` `%` `^` `&` `|` `~` `!` `<` `>` `<<` `>>` `&&` `||` `,` `()` `[]`
+  `+` `-` `*` `/` `%` `^` `&` `|` `~` `!` `<` `>` `<<` `>>` `<=` `>=` `&&` `||` `,` `()` `[]`
 ```
 
 The operators `+`, `-`, `*`, and `&` may only be overloaded in their binary
@@ -652,7 +652,85 @@ take exactly the number of parameters of the corresponding operator as defined
 in this subclause.
 
 ##### Unary Operators
+
+An overloaded prefix unary operator must be implemented by a non-static member
+function with no parameters or a non-member function with one parameter.
+
+Postfix unary operators are not overloadable.
+
 ##### Binary Operators
+
+An overloaded binary operator must be implmeneted by a non-static member
+function with one parameter or a non-member function with two parameters.
+
 ##### Function Call Operator
+
+An overloaded function call operator must be a non-static member function with
+an arbitrary number of parameters, and may have default arguments. Function call
+operator functions are overload candidates for the function call syntax when the
+_postfix-expression_ is an object of class type.
+
 ##### Subscript operator
+
+An overloaded subscript operator must be a non-static member function with one
+parameter. Subscript operator functions are overload candidates for the
+subscript syntax when the _postfix-expression_ is an object of class type.
+
 #### Built-in Operators `[Overload.Builtin]`
+
+// arithmetic
+T operator+(T );
+T operator-(T );
+T operator~(T );
+
+// arithmetic
+LR operator*(L , R );
+LR operator/(L , R );
+LR operator+(L , R );
+LR operator-(L , R );
+bool operator<(L , R );
+bool operator>(L , R );
+bool operator<=(L , R );
+bool operator>=(L , R );
+bool operator==(L , R );
+bool operator!=(L , R );
+LR operator%(L , R );
+
+// enumeration
+bool operator<(T , T );
+bool operator>(T , T );
+bool operator<=(T , T );
+bool operator>=(T , T );
+bool operator==(T , T );
+bool operator!=(T , T );
+
+// integral
+LR operator&(L , R );
+LR operator^(L , R );
+LR operator|(L , R );
+L operator<<(L , R );
+L operator>>(L , R );
+
+// arithmetic
+VQ L & operator=(VQ L &, R );
+VQ L & operator*=(VQ L &, R );
+VQ L & operator/=(VQ L &, R );
+VQ L & operator+=(VQ L &, R );
+VQ L & operator-=(VQ L &, R );
+VQ L & operator%=(VQ L &, R );
+
+// integral
+VQ L & operator<<=(VQ L &, R );
+VQ L & operator>>=(VQ L &, R );
+VQ L & operator&=(VQ L &, R );
+VQ L & operator^=(VQ L &, R );
+VQ L & operator|=(VQ L &, R );
+
+bool operator!(bool);
+bool operator&&(bool, bool);
+bool operator||(bool, bool);
+
+// arithmetic
+LR operator?:(bool, L , R );
+
+operator?:(bool, T , T );
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -137,7 +137,7 @@ declarations cannot be overloaded.
 This section specifies the cases in which a function declaration cannot be
 overloaded. Any program that contains an invalid overload set is ill-formed.
 
- In overload set is invalid if:
+An overload set is invalid if:
   * One or more declaration in the overload set only differ by return type.
 
 ```hlsl
@@ -205,6 +205,13 @@ void J(out int);
 void J(inout int);         // ill-formed: Cannot overload based on out/inout mismatch
 ```
 
+#### Declaration Matching `[Overload.Match]`
+
+Two function declarations with the same name and same parameter lists refer to
+the same function if they are in the same scope (\ref{Basic.Scope}). A function
+declaration in a containing scope is not the same as a function of the same
+name and parameter list in a contained scope.
+
 #### Overload Resolution `[Overload.Res]`
 
 _Overload resolution_ is process by which a function call is mapped to
@@ -231,17 +238,40 @@ candidate set and argument expression list.
 
 ##### Candidate Functions and Argument Lists `[Overload.Res.Sets]`
 
-> \gls{isoCPP} goes into a lot of detail in this section about how candidate
-> functions and argument lists are selected for each context where overload
-> resolution is performed. HLSL matches C++ for the contexts that HLSL inherits.
-> For now, this section will be left as a stub, but HLSL inherits the following
-> sections from C++:
->
-> * **[over.call.func]**
-> * **[over.call.object]**
-> * **[over.match.oper]**
-> * **[over.match.copy]**
-> * **[over.match.conv]**
+At each point where an overload must be resolved an implementation will generate
+a _candidate set_, containing all possible candidate functions, and the argument
+list for the overloaded call. The subclauses under \ref(Overload.Res.Sets)
+define the process by which the candidate set and argument lists are constructed
+in each context where overload resolution occurs.
+
+A candidate set may include both member and non-member functions to be resolved
+against the same argument set. Member functions have an extra _implicit object
+parameter_, representing the object the function is called on. For the purposes
+of overload resolution a static member function is considered to have an
+implicit object parameter.
+
+When appropriate the context shall construct an _implied object argument_
+referring to the object operated on.
+
+The implicit object parameter and implied object argument if present are always
+the first parameter or argument.
+
+The type of the implicit object parameter is: lvalue reference `cv T`; where `T`
+is the class containing the declaration of the function, and `cv` is the
+_cv-qualification_ on the function declaration.
+
+Implied object arguments behave the same as any explicit argument, no temporary
+object shall be created to hold the argument for the implicit object parameter
+except when the implicit object parameter is a member of a
+_cbuffer-declaration_.
+
+If a candidate is a function template, candidate specializations are generated
+through template argument deduction (\ref{Template.Deduction}), candidates are
+then handled as any other candidate. A name may refer to more than one function
+template, in which case candidate functions shall be generated for each function
+template to form the full candidate set.
+
+> TODO: @llvm-beanz to continue fleshing this out...
 
 ##### Viable Functions `[Overload.Res.Viable]`
 
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -662,7 +662,7 @@ Postfix unary operators are not overloadable.
 
 ##### Binary Operators
 
-An overloaded binary operator must be implmeneted by a non-static member
+An overloaded binary operator must be implemented by a non-static member
 function with one parameter or a non-member function with two parameters.
 
 ##### Function Call Operator
--- a/proposals/0003-overload-resolution.md
+++ b/proposals/0003-overload-resolution.md
@@ -118,3 +118,367 @@ non-ambiguous cases.
 This proposal adopts a modified overload resolution algorithm from C++ which is
 updated to account for the new conversions and conversion ranks defined in the
 [implicit conversions](#implicit-conversions) section of this document.
+
+## Detailed Design
+
+### Overloading `[Overload]`
+
+When a single name is declared with two or more different declarations in the
+same scope, the name is _overloaded_. A declaration that declares an
+overloaded name is called an _overloaded declaration_. The set of
+overloaded declarations that declare the same overloaded name are that name's
+_overload set_.
+
+Only function and template declarations can be overloaded; variable and type
+declarations cannot be overloaded.
+
+#### Overloadable Declarations `[Overload.Decl]`
+
+This section specifies the cases in which a function declaration cannot be
+overloaded. Any program that contains an invalid overload set is ill-formed.
+
+ In overload set is invalid if:
+  * One or more declaration in the overload set only differ by return type.
+
+```hlsl
+int Yeet();
+uint Yeet(); // ill-formed: decls differ only by return type
+```
+  * An overload set contains more than one member function declarations with
+  the same _parameter-type-list_, and one of those declarations is a
+  `static` member function declaration (\ref{Classes.Static}).
+```hlsl
+class Doggo {
+  static void pet();
+  void pet();              // ill-formed: static pet has the same parameter-type-list
+  void pet() const;        // ill-formed: static pet has the same parameter-type-list
+
+  void wagTail();          // valid: no conflicting static declaration.
+  void wagTail() const;    // valid: no conflicting static declaration.
+
+  static void bark(Doggo D);
+  void bark();             // valid: static bark parameter-type-list is different
+  void bark() const;       // valid: static bark parameter-type-list is different
+};
+```
+
+  * An overload set contains more than one entry function declaration (\ref{Decl.Attr.Entry}).
+```hlsl
+[shader("vertex")]
+void VS();
+void VS(int);              // valid: only one entry point.
+
+[shader("vertex")]
+void Entry();
+
+[shader("compute")]
+void Entry(int);           // ill-formed: an overload set cannot have more than one entry function
+```
+
+  * An overload set contains more than one function declaration which only
+  differ in parameter declarations of equivalent types.
+```hlsl
+void F(int4 I);
+void F(vector<int, 4> I);  // ill-formed: int4 is a type alias of vector<int, 4>
+```
+
+  * An overload set contains more than one function declaration which only
+  differ in `const` specifiers.
+```hlsl
+void G(int);
+void G(const int);         // ill-formed: redeclaration of G(int)
+void G(int) {}
+void G(const int) {}       // ill-formed: redefinition of G(int)
+```
+
+  * An overload set contains more than one function declaration which only
+  differ in parameters mismatching `out` and `inout`.
+```hlsl
+void H(int);
+void H(in int);            // valid: redeclaration of H(int)
+void H(inout int);         // valid: overloading between in and inout is allowed
+
+void I(in int);
+void I(out int);           // valid: overloading between in and out is allowed
+
+void J(out int);
+void J(inout int);         // ill-formed: Cannot overload based on out/inout mismatch
+```
+
+#### Overload Resolution `[Overload.Res]`
+
+_Overload resolution_ is process by which a function call is mapped to
+a the best overloaded function declaration. Overload resolution uses set of
+functions called the _candidate set_, and a list of expressions that
+comprise the argument list for the call.
+
+Overload resolution selects the function to call in the following
+contexts\footnote{DXC only supports overload resolution for function calls and
+invocation of operators during expressions. Clang will support all contexts
+listed.}:
+
+  * invocation of a function named in a function call expression;
+  * invocation of a function call operator on a class object named in
+  function call syntax;
+  * invocation of the operator referenced in an expression;
+  * invocation of a user-defined conversion for copy-initialization of a
+  class object;
+  * invocation of a conversion function for initialization of an object of a
+  nonclass type from an expression of class type.
+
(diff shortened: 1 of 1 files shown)


bogner (commented):
Typo in PR title (s/langauge/language)

llvm-beanz:
I've moved this back to a draft, I'm going to try and flesh this out a bit more in advance of the next TC57 meeting.

llvm-beanz:
This is now fully fleshed out and ready for review.

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+### Overloading `[Overload]`
+
+When a single name is declared with two or more different declarations in the
+same scope, the name is _overloaded_. A declaration that declares an
+overloaded name is called an _overloaded declaration_. The set of
+overloaded declarations that declare the same overloaded name are that name's
+_overload set_.
+
+Only function and template declarations can be overloaded; variable and type
```
pow2clk:
```suggestion
Only function and function template declarations can be overloaded; variable and type
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+  * An overload set contains more than one function declaration which only
+  differ in parameter declarations of equivalent types.
+```hlsl
+void F(int4 I);
+void F(vector<int, 4> I);  // ill-formed: int4 is a type alias of vector<int, 4>
+```
+
+  * An overload set contains more than one function declaration which only
+  differ in `const` specifiers.
```
pow2clk:
```suggestion
  differ in `const` parameter specifiers.
```

If we are adding const member functions, we'll need to clarify this doesn't apply to them. Even if we aren't, it might be useful

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+  * An overload set contains more than one function declaration which only
+  differ in parameters mismatching `out` and `inout`.
+```hlsl
+void H(int);
+void H(in int);            // valid: redeclaration of H(int)
+void H(inout int);         // valid: overloading between in and inout is allowed
+
+void I(in int);
+void I(out int);           // valid: overloading between in and out is allowed
```
pow2clk, line 202: Neither of the last two are allowed by DXC. I can't imagine how they would be resolved.

https://godbolt.org/z/s6z7bqYsz

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+void J(out int);
+void J(inout int);         // ill-formed: Cannot overload based on out/inout mismatch
+```
+
+#### Declaration Matching `[Overload.Match]`
+
+Two function declarations with the same name and same parameter lists refer to
+the same function if they are in the same scope (\ref{Basic.Scope}). A function
+declaration in a containing scope is not the same as a function of the same
+name and parameter list in a contained scope.
```
pow2clk, line 213: HLSL has inheritance, so some language about member functions in derived classes versus base classes is warranted

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+  the same _parameter-type-list_, and one of those declarations is a
+  `static` member function declaration (\ref{Classes.Static}).
+```hlsl
+class Doggo {
+  static void pet();
+  void pet();              // ill-formed: static pet has the same parameter-type-list
+  void pet() const;        // ill-formed: static pet has the same parameter-type-list
+
+  void wagTail();          // valid: no conflicting static declaration.
+  void wagTail() const;    // valid: no conflicting static declaration.
```
pow2clk: DXC does not accept const members, so this is invalid for a different reason:  https://godbolt.org/z/f84caxY6d

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+    of non-class type from an expression of class type.
+
+In each of these contexts a unique method is used to construct the overload
+candidate set and argument expression list.
+
+##### Candidate Functions and Argument Lists `[Overload.Res.Sets]`
+
+At each point where an overload must be resolved an implementation will generate
+a _candidate set_, containing all possible candidate functions, and the argument
+list for the overloaded call. The subclauses under \ref(Overload.Res.Sets)
```
pow2clk: This is a reference to the current section

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+  * invocation of a user-defined conversion function for initializing an object
+    of non-class type from an expression of class type.
+
+In each of these contexts a unique method is used to construct the overload
+candidate set and argument expression list.
+
+##### Candidate Functions and Argument Lists `[Overload.Res.Sets]`
+
+At each point where an overload must be resolved an implementation will generate
+a _candidate set_, containing all possible candidate functions, and the argument
```
pow2clk: What is the protocol for italics? It seems for the most part that the first introduction of a term is italicized and subsequently not, but this is an exception

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+referring to the object operated on.
+
+The implicit object parameter and implied object argument if present are always
+the first parameter or argument.
+
+The type of the implicit object parameter is: lvalue reference `cv T`; where `T`
+is the class containing the declaration of the function, and `cv` is the
+_cv-qualification_ on the function declaration.
+
+Implied object arguments behave the same as any explicit argument, no temporary
```
pow2clk: This transition makes it sound like that's the case for any explicit argument instead of an exception
```suggestion
Implied object arguments behave the same as any explicit argument except that no temporary
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+expression.
+
+The table below maps unary and binary operator expressions to equivalent member
+and non-member function call expressions. In the table `@` represents any
+spellable operator, and `a` and `b` represent arguments to the operator.
+
+| Expression | Member function   | Non-member function |
+| @a         | (a).operator@()   | operator@(a)        |
+| a@b        | (a).operator@(b)  | operator@(a,b)      |
+| a[b]       | (a).operator[](b) |                     |
```
pow2clk, line 347: The last column assumes the introduction of non-member operator overloads

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+  expression excluding the results from step 1. If no operand is of class type,
+  only functions where `T1` or `T2` are enumerations are added to the set.
+* The built-in operators of `operator@` from the operators defined in
+  \ref{Overload.Builtin} are added to the set.
+
+The left operand to built-in assignment operators, may not have user-defined
+conversion applied to it, and no temporary may be introduced to hold it.
+
+If the operator is the `,` operator, and there are no viable functions, the
+built-in operator is used.
```
pow2clk, line 366: I don't think the comma operator is included in the 2021 documentation, but it does work

pow2clk (commented):
I ran out of gas before I finished here. I'll try to pick it up again, but it might not be until Monday

proposals/0003-overload-resolution.md
llvm-beanz: DXC is complicated. It doesn't allow a user to write this code, but it actually does depend on this for a bunch of the built-in types. All the methods on resource objects and other types generated through hctdb.py and gen_intrin_main.txt are hard coded as `const` (see the `true` here: https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/lib/Sema/SemaHLSL.cpp#L3435).

The valid candidate selection in DXC depends on respecting the const-ness of the `this` object and actually has some hacks to allow functions to be called on user-defined types stored in cbuffers.

I've posted a proposal to allow this syntax to be parsed for user functions (see: https://github.com/hlsl-tc57/tc57/pull/74), but to even match what DXC does today we need the concept of it in the overloading clause.

proposals/0003-overload-resolution.md
llvm-beanz: So, the obnoxious thing is that DXC does allow these to be overridden (the errors are failure to resolve, not illegal overload sets). It actually can resolve the call to the `in` case if you specify constant values.

See: https://godbolt.org/z/Mf4WxMnaM

proposals/0003-overload-resolution.md
llvm-beanz: Class declarations are scopes, so they don't need to be called out separately.

proposals/0003-overload-resolution.md
llvm-beanz:
```suggestion
a candidate set, containing all possible candidate functions, and the argument
```

proposals/0003-overload-resolution.md
llvm-beanz: intentionally so, I guess we could write it as "this clause".

proposals/0003-overload-resolution.md
llvm-beanz: yes

proposals/0003-overload-resolution.md
llvm-beanz: yes, lots of things are not in the HLSL documentation and work.

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+A candidate set may include both member and non-member functions to be resolved
+against the same argument set. Member functions have an extra _implicit object
+parameter_, representing the object the function is called on. For the purposes
+of overload resolution a static member function is considered to have an
+implicit object parameter.
+
+When appropriate the context shall construct an _implied object argument_
+referring to the object operated on.
+
+The implicit object parameter and implied object argument if present are always
```
pow2clk: It's unclear that the "if present" applies to both parameter and argument. It could be moved to the front of the sentence, which would simplify the comma requirements as here it should have commas on both sides, but I think this is cleaner:
```suggestion
Any implicit object parameter or implied object argument is always
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+the first parameter or argument.
+
+The type of the implicit object parameter is: lvalue reference `cv T`; where `T`
+is the class containing the declaration of the function, and `cv` is the
+_cv-qualification_ on the function declaration.
```
pow2clk, line 259: HLSL doesn't have `volatile` at all and doesn't allow `const` on function declarations. If we really think `volatile` is in the future of HLSL, I suppose we could call the const qualifier `cv-qualification`, but there is, unfortunately, currently no way to have a const implicit `this` argument

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+**Function Call Syntax `[Overload.Res.Call]`**
+
+```
+postfix-expression ( \opt{expression-list} )
+```
+
+In a function call (\ref{Expr.Post.Call}), if the _postfix-expression_ names a
+candidate set comprised of overloaded functions and/or function templates, the
+overload resolves as to a named function (\ref{Overload.Res.Call.Named}). If the
+_postfix-expression_ names an object of class type the overload is resolved as a
```
pow2clk:
```suggestion
_postfix-expression_ names an object of class type, the overload is resolved as a
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+forms:
+
+```
+postfix-expression:
+  postfix-expression . id-expression
+  primary-expression
+```
+
+In the first grammar production the _id-expression_ is the name of the function
+to be called, and the _postfix-expression_ preceeding the `.` operator. This
```
pow2clk: If the sentence is otherwise fine, it still needs these changes
```suggestion
In the first grammar production, the _id-expression_ is the name of the function
to be called, and the _postfix-expression_ preceding the `.` operator. This
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+formation represents a _qualified_ function call, where the _postfix-expression_
+must be an object of class type, and the call is looked up as a member function
+of that object. The function names found by that lookup constitute the overload
+candidate set for the call. The argument list for the call is the
+_expression-list_ in the call with the left operand of the `.` operator
+prepended as the implicit object parameter (\ref{Overload.Res.Sets}).
+
+In the second grammar production the name is not qualified by an object
```
pow2clk:
```suggestion
In the second grammar production, the name is not qualified by an object
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+parameter, so it is an unqualified call. Name lookup occurs as normal for name
+lookup of a function call (\ref{Basic.Lookup}), function declarations found
```
pow2clk: Need either a period or semicolon here. I prefer a new sentence
```suggestion
lookup of a function call (\ref{Basic.Lookup}). Function declarations found
```

proposals/0003-overload-resolution.md
pow2clk: This sentence doesn't sound quite right. Should it be 
```suggestion
In the first grammar production, the function call is the _id-expression_, which is the name of the function
to be called, and the _postfix-expression_ preceding the `.` operator. This
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+constitute the candidate set. The candidate set for an unqualified call will be
+comprised either entirely of non-member functions or entirely of member
+functions of some class `T`. If the candidate set is entirely non-member
+functions, the argument list is the _expression-list_. If the candidate set is
+entirely member functions of class `T`, the argument list is the
+_expression-list_ with an implied object argument prepended. If the call
+expression occurs in a context where an implicit object `this` is in scope and
+of class type `T` or a type `T\``, the `this` object will be the implicit object
```
pow2clk, line 314: Was the backtick intended or the more usual prime symbol? It isn't referenced again and I'm unsure if it's meant to reference a class type derived from `T`, which makes sense, or something else I've not thought of. It doesn't specify that the type is a class type, which is confusing.
```suggestion
of class type `T` or a class type `T\'` derived from `T`, the `this` object will be the implicit object
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+and the implicit object argument is a contrived argument, the call is
+ill-formed.
+
+_Call to a Object of Class Type `[Overload.Res.Call.Object]`_
+
+If the _primary-expression_ in the function call expression is a class object of
+type `T`, then the set of candidate functions will include the call operators of
+`T` obtained by name lookup of the name `operator()` (\ref{Basic.Lookup}).
+
+For this formulation the argument list consists of the _expression-list_ from
```
pow2clk:
```suggestion
For this formulation, the argument list consists of the _expression-list_ from
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+the call syntax with the implicit object argument prefixed. The implicit object
+argument in this form is the object produced from the evaluated
+_postfix-expression_.
+
+**Operators in Expressions `[Overload.Res.Oper]`**
+
+Operators in an expression where no operands have a class or enumeration type
+are assumed to be built-in operators and interpreted following \ref{Expr}.
+
+If either operand is of class or enumeration type, overload resolution
```
pow2clk, line 336: Enum operator overloading represents another dependency on a yet-to-be-submitted non-member operator proposal

proposals/0003-overload-resolution.md
pow2clk: Where is that proposal?

proposals/0003-overload-resolution.md
pow2clk: I'm well aware. Are you asking me to refrain from mentioning such cases? Given that we are trying to make sound judgements based on a combination of documentation and compiler behavior, I considered that a relevant fact.

In the meantime, perhaps someone who is able might consider updating the blog or wiki

https://devblogs.microsoft.com/directx/announcing-hlsl-2021/#member-operator-overloading
https://github.com/microsoft/DirectXShaderCompiler/wiki/HLSL-2021

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+**Initialization by Conversion `[Overload.Res.Convert]`**
+
+An object of class type or non-class type `T` may be initialized with an
+initializing expression of class type `S` invoking a user-defined conversion.
+
+The candidate set of conversion functions will be the non-explicit conversion
+functions of `S` and its base classes.
+
+The argument list has one argument which is the initializer expression.
```
pow2clk, line 376: Relies on introduction of conversion operator overloads, another unsubmitted proposal.

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+An overload set is invalid if:
+  * One or more declaration in the overload set only differ by return type.
+
+```hlsl
+int Yeet();
+uint Yeet(); // ill-formed: decls differ only by return type
+```
+  * An overload set contains more than one member function declarations with
+  the same _parameter-type-list_, and one of those declarations is a
+  `static` member function declaration (\ref{Classes.Static}).
```
pow2clk, line 158: There is an awful lot of LaTeX in here

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+##### Best Viable Function `[Overload.Res.Best]`
+
+For an overloaded call with arguments $A_0 ... A_n$, each viable function
+$F(P_0 ... P_m)$, has a set of implicit conversion sequences $ICS_0(F) ... ICS_m(F)$
+defining the conversion sequences for each argument $A_i$ to the
+type of parameter $P_i$.
+
+A viable function $F$ is defined to be a better function than another
+viable function $f\`$ if for all arguments $ICS_i(F)$ is not a worse
```
pow2clk, line 403: More backticks instead of `'`, suggesting it's intentional. At least how github renders it looks pretty strange to me https://github.com/llvm-beanz/tc57/blob/01ee4ca8be6f9960e8c469d60967ef968d7a0e5f/proposals/0003-overload-resolution.md#best-viable-function-overloadresbest

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+required to convert the parameter type back to the argument type
+for writing back to the argument expression lvalue. An inverted implicit
+conversion sequence must be a well-formed implicit conversion sequence where the
+source value is the implicit cxvalue of the parameter type, and the destination
+type is the argument expression's lvalue type.
+
+A well-formed implicit conversion sequence is either a _standard conversion
+sequence_, or a _user-defined conversion sequence_.
+
+In the following contexts an implicit conversion sequence can only be a
```
pow2clk:
```suggestion
In the following contexts, an implicit conversion sequence can only be a
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+| Array-to-pointer | Lvalue Transformation | Exact Match | \ref{Conv.array} |
+| Qualification | Qualification Adjustment | Exact Match | \ref{Conv.qual} |
+| Vector Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
+| Matrix Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.msplat} |
+| Integral promotion | Promotion | Promotion | \ref{Conv.iconv} & \ref{Conv.rank.int} |
+| Floating point promotion | Promotion | Promotion | \ref{Conv.fconv} & \ref{Conv.rank.float} |
+| Component-wise promotion | Promotion | Promotion | \ref{Conv.cwise} |
+| Scalar splat promotion | Scalar Extension Promotion | Promotion Extension | \ref{Conv.vsplat} |
+| Integral conversion | Conversion | Conversion | \ref{Conv.iconv} |
+| Floating point conversion | Convesrion | Conversion | \ref{Conv.fconv} |
```
pow2clk:
```suggestion
| Floating point conversion | Conversion | Conversion | \ref{Conv.fconv} |
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+| Floating-integral conversion | Conversion | Conversion | \ref{Conv.fpint} |
+| Boolean conversion | Convesrion | Conversion | \ref{Conv.bool} |
```
pow2clk:
```suggestion
| Boolean conversion | Conversion | Conversion | \ref{Conv.bool} |
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+| Vector truncation (without conversion) | Dimensionality Reduction | Truncation | \ref{Conv.vtrunc} |
+| Vector truncation promotion | Dimensionality Reduction Promotion | Promotion Truncation | \ref{Conv.vtrunc} |
+| Vector truncation conversion | Dimensionality Reduction Conversion | Conversion Truncation | \ref{Conv.vtrunc} |
+| Matrix truncation (without conversion) | Dimensionality Reduction | Truncation | \ref{Conv.vtrunc} |
+| Matrix truncation promotion | Dimensionality Reduction Promotion | Promotion Truncation | \ref{Conv.vtrunc} |
+| Matrix truncation conversion | Dimensionality Reduction Conversion | Conversion Truncation | \ref{Conv.vtrunc} |
+
+If a scalar splat conversion occurs in a conversion sequence where all other
+conversions are **Exact Match** rank, the conversion is ranked as
+**Extension**. If a scalar splat occurs in a conversion
```
pow2clk, line 530: **Extension** is not a rank listed above, only **Conversion Extension** and **Promotion Extension**

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+Each standard conversion is given a category and rank as defined in the table
+below:
+
+| Conversion | Category | Rank | Reference |
+|------------|----------|------|-----------|
+| No conversion | Identity | Exact Match | |
+| Lvalue-to-rvalue | Lvalue Transformation | Exact Match | \ref{Conv.lval} |
+| Array-to-pointer | Lvalue Transformation | Exact Match | \ref{Conv.array} |
+| Qualification | Qualification Adjustment | Exact Match | \ref{Conv.qual} |
+| Vector Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
```
pow2clk:
```suggestion
| Vector Scalar splat (without conversion) | Scalar Extension | Extension | \ref{Conv.vsplat} |
| Vector Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.vsplat} |
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+| Matrix Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.msplat} |
```
pow2clk:
```suggestion
| Matrix Scalar splat (without conversion) | Scalar Extension | Extension | \ref{Conv.msplat} |
| Matrix Scalar splat conversion | Scalar Extension Conversion | Conversion Extension | \ref{Conv.msplat} |
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+| Integral promotion | Promotion | Promotion | \ref{Conv.iconv} & \ref{Conv.rank.int} |
+| Floating point promotion | Promotion | Promotion | \ref{Conv.fconv} & \ref{Conv.rank.float} |
+| Component-wise promotion | Promotion | Promotion | \ref{Conv.cwise} |
+| Scalar splat promotion | Scalar Extension Promotion | Promotion Extension | \ref{Conv.vsplat} |
```
pow2clk:
```suggestion
| Vector Scalar splat promotion | Scalar Extension Promotion | Promotion Extension | \ref{Conv.vsplat} |
| Matrix Scalar splat promotion | Scalar Extension Promotion | Promotion Extension | \ref{Conv.msplat} |
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+If a scalar splat conversion occurs in a conversion sequence where all other
+conversions are **Exact Match** rank, the conversion is ranked as
+**Extension**. If a scalar splat occurs in a conversion
+sequence with a **Promotion** conversion, the conversion is ranked as
+**Promotion Extension**. If a scalar splat occurs in a conversion
+sequence with a **Conversion** conversion, the conversion is ranked as
+**Conversion Extension**.
+
+If a vector truncation conversion occurs in a conversion sequence where all
```
pow2clk:
```suggestion
If a vector or matrix truncation conversion occurs in a conversion sequence where all
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+other conversions are **Exact Match** rank, the conversion is ranked as
+**Truncation**. If a vector truncation occurs in a conversion
```
pow2clk:
```suggestion
**Truncation**. If a vector or matrix truncation occurs in a conversion
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+sequence with a **Promotion** conversion, the conversion is ranked as
+**Promotion Truncation**. If a vector truncation occurs in a conversion
```
pow2clk:
```suggestion
**Promotion Truncation**. If a vector or matrix truncation occurs in a conversion
```

proposals/0003-overload-resolution.md
pow2clk: It might be preferable to cluster the splats together like the way the truncations are

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+
+For all types `T` where `T` is an integral type
+(\ref{Basic.Types.Arithmetic}); operators will be defined of the form:
+
+```
+LR operator&(L , R );
+LR operator^(L , R );
+LR operator|(L , R );
+L operator<<(L , R );
+L operator>>(L , R );
```
pow2clk, line 727: These compile without diagnostics for floats just as the `~` and `%` operators do. I'm not sure that should be considered as being implicitly defined for those types, but if these are limited to integers, `~` and `%` should be too.

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +126,654 @@ non-ambiguous cases.
...
+the ternary operator will be defined of the form:
+
+```
+LR operator?:(bool, L , R );
+```
+
+For all types `T`, a ternary operator will be defined of the form:
+
+```
+operator?:(bool, T , T );
```
pow2clk:
```suggestion
T operator?:(bool, T , T );
```

proposals/0003-overload-resolution.md
pow2clk: With that proposal listed as a dependency, I'll be mollified, but this is quite clearly a user-defined class

proposals/0003-overload-resolution.md
pow2clk: Literals clearly can't match the out parameter, so there is no ambiguity.  How in any compiler would a non-const variable argument be resolved? If it would just go to the `out` or `inout` version, that would be consistent, but I expect would likely catch people unawares. I think this should be ambiguous and refuse to resolve.

proposals/0003-overload-resolution.md
pow2clk: The C++ spec disagrees and so do I https://timsong-cpp.github.io/cppwp/n3337/over#dcl

proposals/0003-overload-resolution.md
pow2clk: I would prefer that

pow2clk (commented):
I've counted a lot of dependencies here, most on proposals that don't yet exist at least not in this repo. I think they should exist and be referenced before this goes in

proposals/0003-overload-resolution.md
llvm-beanz: Are you requesting a change? I left the `ref` markers because there isn't a reasonable way that I could think of to indicate and preserve references to other sections in the specification.

proposals/0003-overload-resolution.md
llvm-beanz: Roughly equivalent code in C++ is not ambiguous so I'm not sure I see why this should be. Consider:

```c++

float H(int c) {return 1.0;}
float H(int &c) { c = 0; return 2.0;}

float HH(int c) {return 1.0;}
float HH(int &c) { c = 0; return 2.0;}

float fn(float f, int ct) {
    float res = H(1);
    res += HH(2);
    return res + ct;
}
```
[Compiler Explorer](https://godbolt.org/z/PE48hdzox)

proposals/0003-overload-resolution.md
llvm-beanz: The C++ spec does not disagree with my point, it actually states explicitly exactly what I said:

> A function member of a derived class is _not_ in the same scope as a function member of the same name in a base class.

A function declared in a base class, and a function of the same name and parameter list declared in a derived class are defined in different scopes, so they are different functions. I did not feel it was necessary to re-state that derived classes and base classes are different scopes.

Do you have suggested language that you would like to see included here?

proposals/0003-overload-resolution.md
llvm-beanz:
```suggestion
list for the overloaded call. The subclauses under this clause
```

proposals/0003-overload-resolution.md
llvm-beanz: There is no way in a user-defined function to have a const-qualified `this`, however many of the built-in data types do.

Let me separate out two different issues:

1) I've chosen to use the grammar `cv-qualification` and that notation to align with the C++ definition with no intention of adding the `volatile` keyword. `cv-qualification` is already in use in the HLSL spec language we've adopted. If you think we should change that please file an issue, but for now this proposal is consistent with the existing spec, and we shouldn't change it.

2) I understand (although disagree) with your point above about the example spellings, but regardless of whether or not the cv-qualification is allowed to be spelled for a member function declaration (that is a separate proposal), I think the overload resolution rules need to be adapted to address this.

Consider the following code that DXC compiles "successfully":

```hlsl
Texture2D A;

struct Tex {
  int Loads;
  Texture2D B;

  float4 Load(uint Idx) {
    Loads += 1;
    return B.Load(Idx);
  }
};

float4 GetTex(const inout Tex T) {
    return T.Load(0);
}

float4 Get(const Texture2D T) {
    return T.Load(0);
}

float4 main() : SV_Target {
    Tex T = {0, A};
    float4 Result = GetTex(T) + Get(A);
    return Result + T.Loads;
}
```
[Compiler Explorer](https://godbolt.org/z/x15qfGhjM)

This code illustrates two things: (1) that calling functions on const objects is required for DXC, and (2) that DXC currently does so in a way that violates the const-ness of the objects.

The question that we need to answer is "do we think DXC is doing the right thing, or do we need to fix it?"

The proposal here is that DXC's built-in types (which already have const-qualified methods) should be respected, and we should by these rules disallow resolving a user-defined member function against a `const` object unless we also adopt the proposal in https://github.com/hlsl-tc57/tc57/pull/74.

proposals/0003-overload-resolution.md
llvm-beanz: https://github.com/hlsl-tc57/tc57/pull/115

proposals/0003-overload-resolution.md
llvm-beanz: Is there a specific change you're requesting to the spec? Mentioning that it is different from the documentation but matches the implementation isn't actionable.

This is not the right forum to request that Microsoft update their blog posts or other documentation.

proposals/0003-overload-resolution.md
llvm-beanz: Conversion operators is not a different proposal, that was always part of this proposal. They were intended and expected to work when HLSL 2021 released, but didn't.

proposals/0003-overload-resolution.md
llvm-beanz: They are listed in their ranked order.

proposals/0003-overload-resolution.md
llvm-beanz: No, the function call is the whole expression. This is talking about the _name_ of the function as used for overload resolution.

proposals/0003-overload-resolution.md
llvm-beanz: I didn't care to want to use a proper prime symbol in the markdown because it's not a standard keyboard symbol and I assumed people could get enough context using a backtick.

Your suggestion didn't remove the escape, so unless you feel strongly about it I'll ignore this.

proposals/0003-overload-resolution.md
llvm-beanz: So, the intent of this table is to render with groupings of categories and ranks. There are some errors here specifically with the matrix elements repeating instead of merging rows in the table.

Screenshot 2026-06-23 at 9 10 01 AM

The goal is to group conversions by categories and ranks to simplify. Markdown tables aren't as flexible as LaTeX, so some of this gets lost.

proposals/0003-overload-resolution.md
llvm-beanz: Do you have an example? They seem to fail for me https://godbolt.org/z/Kq9qqW8on.

llvm-beanz:
@pow2clk I believe I've addressed all your feedback. Please let me know if there is anything else you want addressed before merging this.

Reminder: This is _not_ final normative language, so it doesn't need to wait until the next meeting to merge. I'd like to get this merged this week if possible.

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +127,657 @@ non-ambiguous cases.
...
+  void pet();              // ill-formed: static pet has the same parameter-type-list
+  void pet() const;        // ill-formed: static pet has the same parameter-type-list
+
+  void wagTail();          // valid: no conflicting static declaration.
+  void wagTail() const;    // valid: no conflicting static declaration.
+
+  static void bark(Doggo D);
+  void bark();             // valid: static bark parameter-type-list is different
+  void bark() const;       // valid: static bark parameter-type-list is different
+};
```
bogner, line 169: I don't know that the const overloads are really adding much here, and the fact that each example has two overloads that aren't static muddies up what this is trying to say. I'd probably go more along these lines:
```suggestion
class C {
  static void f1();
  void f1();           // ill-formed: cannot be overloaded with `static void f1()`

  static void f2();
  void f2(int);        // valid

  void f3(int);
  static void f3(int); // ill-formed: cannot be overloaded with `void f3(int)`
};
```

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +127,657 @@ non-ambiguous cases.
...
+[shader("compute")]
+void Entry(int);           // ill-formed: an overload set cannot have more than one entry function
+```
+
+  * An overload set contains more than one function declaration which only
+  differ in parameter declarations of equivalent types.
+```hlsl
+void F(int4 I);
+void F(vector<int, 4> I);  // ill-formed: int4 is a type alias of vector<int, 4>
+```
```
bogner: This doesn't seem right - this is just two declarations with the same type. Since they're not definitions, this is the same as if we do:
```hlsl
void F(int);
void F(int);
```
which seems perfectly legal to me. The related case is also not really worth including text for IMO:
```hlsl
void F(int) {}
void F(int) {} // error: redeclaration
```
this isn't an overload at all, it's just a redeclaration, which is invalid for reasons completely unrelated to overloads.

proposals/0003-overload-resolution.md
bogner: FWIW I don't necessarily agree that the C++ example here is roughly equivalent. `int` and `int &` are distinct types (that can be converted between in certain situations). `int` and `in int` are certainly not distinct types. I think this is more akin to `int`/`const int` than it is to references - it's semantically less similar sure, but I mean this in terms of how these qualifiers relate to types. I can see how making this unambiguous is implementable, but I personally think it isn't very useful.

Is there some motivation for allowing `in int`/`out int` conflicts, or are we just trying to make this unambiguous because we can?

proposals/0003-overload-resolution.md
```diff
@@ -118,3 +127,657 @@ non-ambiguous cases.
...
+##### Unary Operators
+
+An overloaded prefix unary operator must be implemented by a non-static member
+function with no parameters or a non-member function with one parameter.
+
+Postfix unary operators are not overloadable.
+
+##### Binary Operators
+
+An overloaded binary operator must be implmeneted by a non-static member
```
bogner:
```suggestion
An overloaded binary operator must be implemented by a non-static member
```

bogner (commented):
Generally looks good other than a few minor comments and some whinging about how we should deal with some of the nuances of `in/out/inout`

proposals/0003-overload-resolution.md
llvm-beanz: I started writing this from the perspective of what works today. Today in both DXC and Clang `out` and `inout` modify the type of the parameter effectively to something reference-like (it is a reference both ASTs), which allows overloading based on `out` or `inout`.

See:
```hlsl
struct Doggo {
    void SetWagging(bool B) {
        Wagging = B;
    }

    void SetWagging(inout bool B) {
        Wagging = B;
        B = !B;
    }

    bool Wagging;
};

[numthreads(1,1,1)]
void main() {
    bool B = true;
    Doggo D = {false};
    D.SetWagging(true); // Works on Clang, ambiguous on DXC
    D.SetWagging(B); // Ambiguous everywhere
}
```
[Compiler Explorer](https://godbolt.org/z/Mqdfrzof8)

There is a very small difference in behavior today between Clang and DXC in that Clang will resolve the call if the argument is an rvalue-reference because it excludes `inout` as a candidate because you cannot bind an rvalue expression to an `inout` parameter. There is an bug in DXC here because DXC includes the `inout` function in the overload candidate set even though it isn't viable, and it rejects it later in the compiler when binding the parameter to the selected best match. Once you've selected a matching function you shouldn't be able to fail binding parameters, that can only happen if you selected a non-matching function.

```hlsl
bool Nonsense(inout bool B) {
    B = !B;
}

[numthreads(1,1,1)]
void main() {
    Nonsense(true);
}
```
[Compiler Explorer](https://godbolt.org/z/cMMKhn7cc)

If we want to change the behavior of the language to make overloading based on `inout` and `out` illegal we can do that. It probably won't break existing code (at least not code that is actually in use), but it is a change in behavior from what all existing compilers do today.

bogner: Ah, okay, I see how this behaviour falls out of how we've defined `cxvalue` in the spec now. From the perspective of the function definition (both in terms of dxc and clang's implementations but also as part of how we've specified the `cxvalue` idea) we essentially have a reference here. So in terms of implementing the overload set this is in fact pretty similar to overloads of references. The fact that there needs to be a conversion at the call site is what makes us able to differentiate between `lvalue` and `rvalue` here and do the right thing.

Given that this is kind of the natural way things fall out of the rest of the language spec I'm fairly convinced what you have here is the better path forward. Arbitrarily banning the conversion is feasible but actually adds some complexity for little value.

bogner (approved):

llvm-beanz:
@pow2clk are you satisfied with this as it is, or are there things you think still need to be updated?

llvm-beanz:
@pow2clk, if you have additional changes you'd like to see to this language please file issues or create a PR and we can work through that. It will be a while before this proposal comes up for stage advancement so there is still lots of time to adjust this.

proposals/0003-overload-resolution.md
pow2clk: I was trying to explain why I thought the point I was making was relevant. Your response was entirely dismissive of it. I'm not arguing that we should remove support for the comma operator, but that its absence from any documentation at present should make that part of the conversation since we will have many instances of things that are absent from documentation and present in the implementation. Some of them we will decide to keep as the implementation provides. Others we may not. I thought perhaps you would be interested in updating documentation to eliminate this conflict, but it seems we have not even had the conversation, so I'll ask directly:

Are we including the comma operator? I think we should. You may choose to make documentation reflect that. It is outside this forum, but it would be helpful presuming we go the way I suggest.

llvm-beanz: I apologize if my comment came off as dismissive. You've left comments that aren't actionable (at least not in the context of TC57), and I've been genuinely unsure what you're asking for.

I don't see any reason why we would exclude the comma operator, which is why I included it.

PR #63, opened 2026-06-02
[Basic.Scope] Add Scope section to Basic clause

### LINKED ISSUE #94: [Basic.Scope] Scope
We need to draft specification language to define the concept of scopes as it relates to the ranges of programs where declarations are valid.

llvm-beanz:
This adds language to describe the concept of declaration scopes which is necessary for further describing name lookup, and overloading.

Resolves #94

### DIFF
--- a/spec/basic.tex
+++ b/spec/basic.tex
@@ -119,6 +119,31 @@ tbuffer TB {                 // does not declare TB
 
 \Sec{Scope}{Basic.Scope}
 
+\p A \textit{valid} name is a name that may be used unqualified to refer to a
+specific entity.
+
+\p The \textit{declarative region} of a declaration is the largest part of a
+program in which the name is valid.
+
+\p The \textit{scope} of a name is the possibly discontiguous region of a
+program in which the name may be used as an unqualified name to refer to the
+same entity. The scope of a declaration is the same as its declarative region,
+unless another declaration of the same name exists within that region, in which
+case the declarative region of the inner declaration is excluded from the outer
+declaration's scope.
+
+\p Names declared by a declaration are introduced into the scope in which the
+declaration occurs except for members of \textit{cbuffer-declarations}
+(\ref{Decl.cbuffer}), and \textit{using-declarations} (\ref{Decl.Using}) which
+do not follow this general behavior.
+
+\p Within a declarative region, if multiple declarations introduce the same name
+either all the declarations shall refer to the same entity, or all refer to
+functions or function templates; or one declaration shall declare a class or
+enumeration name, and the other declarations shall refer to the same entity, or
+all refer to functions or function templates, hiding the class or enumeration
+name.
+
 \Sec{Name Lookup}{Basic.Lookup}
 
 \Sec{Storage Duration}{Basic.Storage}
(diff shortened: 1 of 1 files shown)


spec/basic.tex
```diff
@@ -119,6 +119,34 @@
 
 \Sec{Scope}{Basic.Scope}
 
+\p A \textit{valid} name is a name that may be used unqualified to refer to a
```
pow2clk, line 122: Does "unqualified" refer to the namespace or struct scope prefixes which would constitute qualifiers?

spec/basic.tex
```diff
@@ -119,6 +119,34 @@
...
+specific entity.
+
+\p The \textit{declarative region} of a declaration is the largest part of a
+program in which the name is valid.
+
+\p The \textit{scope} of a name is the possibly discontiguous region of a
+program in which the name may be used as an unqualified name to refer to the
+same entity. The scope of a declaration is the same as its declarative region,
```
pow2clk:
```suggestion
same entity. The scope of a declaration is the same as its declarative region
```

spec/basic.tex
```diff
@@ -119,6 +119,34 @@
...
+declaration occurs except for members of \textit{cbuffer-declarations}
+(\ref{Decl.cbuffer}), and \textit{using-declarations} (\ref{Decl.Using}) which
+do not follow this general behavior.
+
+\p Within a declarative region, if multiple declarations introduce the same name
+either all the declarations shall refer to the same entity, or all refer to
+functions or function templates; or one declaration shall declare a class or
+enumeration name, and the other declarations shall refer to the same entity, or
+all refer to functions or function templates, hiding the class or enumeration
+name.
```
pow2clk: This is quite the complicated sentence. How many possibilities are we talking about here? By my count:
1. All declarations refer to the same entity (lacking externs, I'm not sure how this would happen for something not a function, class, or enum)
2. All refer to functions or function templates (duplicate function declarations)
3. One declares a class or enum and all others refer to a class or enumeration name and the other declarations refer to that same entity (duplicate class/enum declarations)
4. One declares a class or enum and all others refer to functions or function templates (A class or enum gets silently overridden by the declaration of a function)

I think it would be clearer to just enumerate what possibilities we have in a list

spec/basic.tex
```diff
@@ -119,6 +119,34 @@
...
+
+\p Scopes nest within eachother. A \textit{containing scope}, is a scope that
```
pow2clk: typo
```suggestion
\p Scopes nest within each other. A \textit{containing scope}, is a scope that
```

spec/basic.tex
llvm-beanz: I haven't written the name lookup section yet, but our name resolution rules should match C++ identically since I don't think we have any name-lookup specific changes in DXC.

A name that is used `unqualified` would be a name without a prefixing scope specifier (i.e. a namespace name, type name, or `::`).

spec/basic.tex
pow2clk: Happy with the new text

pow2clk (approved):

llvm-beanz:

bogner (approved):

PR #64, opened 2026-06-02
[AnnexB] Add informative Annex B

llvm-beanz:
This is where we will document all the skeletons we hid in the closet...

### DIFF
--- a/spec/annex-b.tex
+++ b/spec/annex-b.tex
@@ -1,0 +1,27 @@
+\Annex{informative}{Implementation Documentation}{AnnexB}
+
+\Sec{Implementation Limits}{AnnexB.Limits}
+
+\p This section describes reasonable implementation limits that may be assumed
+by a conforming implementation. These limits specify minimum required or maximum
+allowed values for various implementation-defined parameters.
+
+\p \ref{Basic.Types} declares a vector type as having required supported sizes
+of 1 to 4 components. Implementations may support larger vector sizes but are
+not required to do so. An implementation that supports larger vector sizes must
+define its vector size limit.
+
+\Sec{Observed Deviations}{AnnexB.Documentation}
+
+\p This section provides documentation for users around behavior observed by the
+specification authors around specific implementations. This documentation is
+provided for historical context and information purposes only.
+
+\Sub{Resource Objects}{AnnexB.Resources}
+
+\p In violation of \ref{Basic.Types.Intangible}, DXC supports storing objects
+that contain resources in \texttt{cbuffer} declarations. In this use case the
+resource objects are not actually stored in the \texttt{cbuffer}, but instead
+the resource is treated as a global resource declaration and name lookup
+resolves the value inside the structure declaration to the implied global
+resource declaration.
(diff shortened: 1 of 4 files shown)


spec/annex-b.tex
```diff
@@ -0,0 +1,27 @@
+\Annex{informative}{Implementation Documentation}{AnnexB}
+
+\Sec{Implementation Limits}{AnnexB.Limits}
+
+\p This section describes reasonable implementation limits that may be assumed
+by a conforming implementation. These limits specify minimum required or maximum
+allowed values for various implementation-defined parameters.
```
pow2clk, line 64: Is this intended to establish hard requirements or just give a sense of what kind of limits can be expected? The "informative" part and that it's an annex suggests the latter, but words like "assumed by a conforming implementation" sound more normative"

pow2clk (commented):
As much as I like the colorful description, I don't think it explains what the change actually does

spec/annex-b.tex
llvm-beanz: I followed the pattern with the C++ specification where implementation limits are informative.

The normative text doesn't define the limits, so the expectation is that they would be infinite, but the informative annex provides more reasonable limits acknowledging that computers don't have infinite memory and that implementations may need to have reasonable limits for performance or even just possibly working.

jessebarker (approved):
As an informative section, I'm in favor of this.  As discussed on the call, it doesn't have to be exhaustive, but it seems like we could capture details as other spec PRs are reviewed and land.

gesa:
Informative note to make brain gears go: perhaps this or another Annex may also contain a clause for documenting invariants, for example when something winds up as kebab case despite similar attributes being snake case.

Proof of concept: https://github.com/codehag/documenting-invariants

spec/annex-b.tex
```diff
@@ -0,0 +1,27 @@
...
+\p \ref{Basic.Types} declares a vector type as having required supported sizes
+of 1 to 4 components. Implementations may support larger vector sizes but are
+not required to do so. An implementation that supports larger vector sizes must
+define its vector size limit.
+
+\Sec{Observed Deviations}{AnnexB.Documentation}
+
+\p This section provides documentation for users around behavior observed by the
+specification authors around specific implementations. This documentation is
+provided for historical context and information purposes only.
```
bogner: I think we want to be a bit clearer about what the purpose of documenting these behaviours is. Just mentioning "observed behaviors" doesn't give a good picture of why we've chosen these particular things to discuss here. I might even go so far as to title the section something like "Known Deviations, Breaking Changes, and Gotchas".

Here's an attempt at capturing this:
```suggestion
\p This section provides documentation around implementation behaviors that are
known to differ between specific implementations. This includes breaking
changes between historical HLSL and standard HLSL where some compilers or
versions of compilers may not conform to standard HLSL, as well as observed
non-conformant behaviors in such compilers. This is not meant to be a
comprehensive list of bugs.
```

I also think there would be some value to defining the compilers we discuss below. That is, some sort of glossary-like definitions to explain what compiler we're referring to when we say DXC, FXC, or Clang.

bogner (approved):

Alexander-Johnston (approved):

pow2clk:
> As much as I like the colorful description, I don't think it explains what the change actually does

I would have preferred this addressed. The PR description does not describe the purpose of the section. Looking through history in the future will be unaided by this description.

PR #65, opened 2026-06-02
[Resources] Add resources clause and typed buffers

llvm-beanz:
This adds the typed buffer section of the resources clause. There are some unresolved references to other sections in the resources clause that are yet to be filled out, but this is a good starting point.

### DIFF
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -9,7 +9,10 @@ objects.
 \p Some of the constructs in the standard library are not expressible with
 spellable syntax. This section describes syntax constructs which are
 non-standard but are used throughout the standard library specification.
-All code blocks in the library clauses are non-normative.
+
+\p Code blocks in the remaining clauses of the specification decribe a
+hypothetical implementation and are provided for clarity. A conforming
+implementation need not rely on the specifc code in any of the blocks. 
 
 \Sub{Constructors}{Library.Notation.Constructors}
 
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -1,0 +1,31 @@
+\Ch{Library Overview}{Library}
+
+\p This clause serves as an overview to the remaining clauses of the
+specification which detail \acrshort{hlsl}'s standard library of functions and
+objects.
+
+\Sec{Syntax Notations}{Library.Notation}
+
+\p Some of the constructs in the standard library are not expressible with
+spellable syntax. This section describes syntax constructs which are
+non-standard but are used throughout the standard library specification.
+
+\Sub{Constructors}{Library.Notation.Constructors}
+
+\p Some standard library types have behaviors associated in \gls{isoCPP} with
+constructors. These syntaxes are intended to describe compatible behavior with
+C++ constructors.
+
+\Sub{Reference Types}{Library.Notation.References}
+
+\p Some standard library functions or member functions behave as if a parameter
+or return value are references (passed by address or returned by address). To
+simplify this notation the syntax \textit{address-space\textsubscript{opt}}
+\texttt{T \&} is used where the optional \textit{address-space} is one of
+\texttt{device} or \texttt{groupshared}.
+
+\Sub{Concepts}{Library.Notation.Concepts}
+
+\p Some standard library template functions or objects have constraints to which
+types are valid as template arguments. To simplify this notation the concepts
+syntax from \gls{isoCPP20} is used.
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -79,11 +79,11 @@ of any introduced headers, macros, and declarations.
 that apply:
 
 \begin{itemize}
-  \item name and description
-  \item class definition or function prototype
-  \item template argument constraints
-  \item class invariants
-  \item function semantics
+  \item Name and description
+  \item Class definition or function prototype
+  \item Template argument constraints
+  \item Class invariants
+  \item Function semantics
 \end{itemize}
 
 \p Description of class member functions will follow the order, omitting
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -12,7 +12,7 @@ non-standard but are used throughout the standard library specification.
 
 \p Code blocks in the remaining clauses of the specification decribe a
 hypothetical implementation and are provided for clarity. A conforming
-implementation need not rely on the specifc code in any of the blocks. 
+implementation need not rely on the specifc code in any of the blocks.
 
 \Sub{Constructors}{Library.Notation.Constructors}
 
@@ -33,3 +33,80 @@ simplify this notation the syntax \textit{address-space\textsubscript{opt}}
 \p Some standard library template functions or objects have constraints to which
 types are valid as template arguments. To simplify this notation the concepts
 syntax from \gls{isoCPP20} is used.
+
+\Sec{Method of Description (Informative)}{Library.Method}
+
+\Sub{Clause Structure}{Libarry.Method.Structure}
+
+\p Each library clause will contain all applicable subclauses of the list:
+
+\begin{itemize}
+  \item Introduction
+  \item Optional features
+  \item Special semantics
+  \item Element subclauses
+\end{itemize}
+
+\p The clause introduction shall provide an overview of the contents of the
+clause. Some features detailed in a clause may be optional, a subclause
+describing the optional features will be included if applicable.
+
+\p Each library clause will optionally contain a subclause describing additional
+semantic behaviors required for the clause. This subclause is omitted if empty.
+
+\p Each library clause will contain one or more element subclauses detailing the
+declarations required in a conforming implementation of the specification, and
+any normative optional declarations based on optional features described in the
+clause.
+
+\SubSub{Element Subclauses}{Library.Method.Structure.Element}
+
+\p Each element subclause contain the following:
+
+\begin{itemize}
+  \item Summary
+  \item Detailed specifications
+\end{itemize}
+
+\SubSub{Summary}{Library.Method.Structure.Summary}
+
+\p The summary will include a description of the category as well as a listing
+of any introduced headers, macros, and declarations.
+
+\SubSub{Detailed Specification}{Library.Method.Structure.Detailed}
+
+\p The detailed specification for each entity will contain each of the below
+that apply:
+
+\begin{itemize}
+  \item name and description
+  \item class definition or function prototype
+  \item template argument constraints
+  \item class invariants
+  \item function semantics
+\end{itemize}
+
+\p Description of class member functions will follow the order, omitting
+sections that are empty:
+
+\begin{itemize}
+  \item Constructors
+  \item Copying and assignment
+  \item Modifying functions
+  \item Inspection functions
+  \item Operators
+  \item Other non-member functions
+\end{itemize}
+
+\p Descriptions of function semantics will contain all of the following elements
+that apply in the order as specified:
+
+\begin{itemize}
+  \item \textit{Requires:} the preconditions for calling the function.
+  \item \textit{Effects:} the actions performed by the function.
+  \item \textit{Synchronization:} the requirements for execution or memory
+  synchronization.
+  \item \textit{Returns:} a description of the value returned by the function.
+  \item \textit{Remarks:} additional normative information about the function.
+  \item \textit{Notes:} non-normative information provided about the function.
+\end{itemize}
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -9,6 +9,7 @@ objects.
 \p Some of the constructs in the standard library are not expressible with
 spellable syntax. This section describes syntax constructs which are
 non-standard but are used throughout the standard library specification.
+All code blocks in the library clauses are non-normative.
 
 \Sub{Constructors}{Library.Notation.Constructors}
 
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -12,7 +12,7 @@ non-standard but are used throughout the standard library specification.
 
 \p Code blocks in the remaining clauses of the specification decribe a
 hypothetical implementation and are provided for clarity. A conforming
-implementation need not rely on the specifc code in any of the blocks.
+implementation need not rely on the specific code in any of the blocks.
 
 \Sub{Constructors}{Library.Notation.Constructors}
 
--- a/spec/library.tex
+++ b/spec/library.tex
@@ -13,7 +13,7 @@ non-standard but are used throughout the standard library specification.
 \Sub{Constructors}{Library.Notation.Constructors}
 
 \p Some standard library types have behaviors associated in \gls{isoCPP} with
-constructors. These syntaxes are intended to describe compatible behavior with
+constructors. These syntaxes are intended to describe equivalent behavior with
 C++ constructors.
 
 \Sub{Reference Types}{Library.Notation.References}
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -157,7 +157,7 @@ T Load(uint index) const;
 \end{HLSL}
 
 \begin{itemize}
-  \item \textit{Requires:} \texttt{index} < the number of elements in the
+  \item \textit{Requires:} \texttt{index} \textless the number of elements in the
   typed buffer.
   \item \textit{Returns:} The element in the typed buffer at the index specified
   by \texttt{index}.
@@ -188,7 +188,7 @@ T operator[](uint index) const;
 \end{HLSL}
 
 \begin{itemize}
-  \item \textit{Requires:} \texttt{index} < the number of elements in the
+  \item \textit{Requires:} \texttt{index} \textless the number of elements in the
   typed buffer.
   \item \textit{Returns:} The element in the typed buffer at the index specified
   by \texttt{index}.
@@ -199,7 +199,7 @@ device T &operator[](int index);
 \end{HLSL}
 
 \begin{itemize}
-  \item \textit{Preconditions:} \texttt{index} < the number of elements in the
+  \item \textit{Preconditions:} \texttt{index} \textless the number of elements in the
   typed buffer.
   \item \textit{Returns:} An lvalue-reference to the element in the typed buffer
   at the index specified by \texttt{index}.
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -1,6 +1,7 @@
 \Ch{Resource Types}{Resources}
 
-\p Resource types are intangible types representing memory or configuration with an external interface.
+\p Resource types are intangible types representing memory or configuration with
+an external interface.
 
 \p These take the form of typed buffers, raw buffers, textures, constant buffers
 and samplers.
@@ -36,7 +37,7 @@ subscript indexing into the \texttt{ResourceDescriptorHeap}
 \p A resource declaration in a global scope are called \textit{global
 resource declaration} and represents a binding for external data into a
 program. Global resource declarations are implicitly initialized in an
-implementation-defined.
+implementation-defined way.
 
 \p Use of an uninitialized resource is undefined.
 
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -1,0 +1,177 @@
+\Ch{Resource Types}{Resources}
+
+\p Resources are intangible types representing memory with an external interface.
+
+\p These take the form of Typed Buffers, Raw Buffers, Textures, Constant Buffers
+and Samplers.
+
+\p Typed Buffer, Raw Buffer, and Texture resources may be read-only or
+read-write as denoted by the type of the resource, where the names of read-write
+resources are prefixed with \texttt{RW}.
+
+\begin{note}
+The resource class objects in this clause require language behaviors not
+defined in the current language. Specifically, they utilize constructor
+behaviors, reference types, and concepts (for constraining instantiation). For
+ease of reading this clause uses C++ syntax to represent those language
+constructs. Additionally this clause assumes the introduction of the
+\texttt{device} keyword to denote references to device memory.
+\end{note}
+
+\p Global resource declarations are \textit{bound} to device memory managed by
+an application through a runtime implementation. A resource declaration that is
+not bound is said to be \textit{unbound}. Operations on unbound resources are
+evaluated as no-ops.
+
+\p An implementation may impose restrictions on how local resource objects may
+be used (\ref{AnnexB.LocalResources}).
+
+\Sec{Typed Buffers}{Resources.TypedBuffers}
+
+\p The typed buffer class template represents a one-dimensional resource
+containing an array of a single given type. Its contents are indexed by typed
+access to each element. Types may have their formats converted upon load in an
+implementation-defined way.
+
+\p The element type may be any type up to 16 bytes in size. Elements may be
+padded to 16 bytes if the element type is smaller than 16 bytes.
+
+\p All typed buffers can be read through subscript operators or Load methods.
+Writable typed buffers can be written through subscript operators.
+
+\p Where \texttt{T} can be any arithmetic type \ref{Basic.Types.Arithmetic}
+or a vector with a maximum of four elements containing such an arithmetic type.
+The total size of a single contained element must be less than 16 bytes.
+
+\p Typed buffers perform format conversions on load such that the underlying
+data gets converted to the destination type in an implementation-defined way.
+
+\begin{HLSL}
+template <typename T = float4> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
+class Buffer {
+  Buffer();
+  Buffer(const Buffer &buf);
+  Buffer &operator=(Buffer &buf);
+  Buffer(const __DynamicResourceHandle &handle);
+
+  void GetDimensions(out uint width) const;
+
+  // Element access.
+  T operator[](int index) const;
+  T Load(int index) const;
+  T Load(int index, out uint status) const;
+};
+
+
+template <typename T> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
+class RWBuffer {
+  RWBuffer();
+  RWBuffer(RWBuffer buf);
+  RWBuffer &operator=(RWBuffer &buf);
+  Buffer(const __DynamicResourceHandle &handle);
+
+  void GetDimensions(out uint width) const;
+
+  // Element access.
+  T operator[](int index) const;
+  device T &operator[](int n) const;
+  T Load(int index) const;
+  T Load(int index, out uint status) const;
+};
+\end{HLSL}
+
+\Sub{Constructors}{Resources.TypedBuffers.Ctrs}
+
+\p When defined at global scope, typed buffers are bound to externally-defined
+backing stores using the explicit binding location if provided or the implicit
+binding if not (\ref{Resources.Binding}).
+
+\p When defined at local scope, typed buffers represent local references that
+can be associated with global buffers when assigned. Use of an uninitialized
+typed buffer is undefined.
+
+\begin{HLSL}
+Buffer<int> grobuf;
+RWBuffer<int> grwbuf;
+void main() {
+  Buffer<int> lrobuf = grobuf;
+  RWBuffer<int> lrwbuf = grwbuf;
+}
+\end{HLSL}
+
+\p Buffers cannot be implicitly nor explicitly cast from one type to another.
+
+\p Buffers may be created from dynamic resources. The
+\texttt{\_\_DynamicResourceHandle} type in the examples above is a placeholder
+for the implementation-specific type returned by subscript indexing into the
+\texttt{ResourceDescriptorHeap} (\ref{Resources.Binding.Dynamic}).
+
+\Sub{Dimensions}{Resources.TypedBuffers.Dims}
+
+\p The size of a typed buffer can be retrieved using the \texttt{GetDimensions}
+method.
+
+\begin{HLSL}
+void GetDimensions(out uint width) const;
+\end{HLSL}
+
+\p This returns the full length, in elements of the buffer through the
+\texttt{width} output parameter.
+
+\Sub{Element Access}{Resources.TypedBuffers.Access}
+
+\p The contents of typed buffers can be retrieved using the \texttt{Load} methods
+or the subscript operator.
+
+\begin{HLSL}
+T Load(in int index) const;
+T operator[](int index) const;
+\end{HLSL}
+
+\p These each return the element of the buffer's type at the given index
+\texttt{index}.
+
+\p An additional \texttt{Load} method returns the indicated element just as the
+other, but also takes a second \texttt{status} parameter that returns
+information about the accessed resource.
+
+\begin{HLSL}
+T Load(in int index, out uint status) const;
+\end{HLSL}
+
+\p The \texttt{status} parameter returns an indication of whether all of the
+retrieved value came from fully mapped parts of the resource. This parameter
+must be passed to the built-in function \texttt{CheckAccessFullyMapped}
+(\ref{Resources.Mapcheck}) in order to interpret it.
+
+\p Writable buffers have an additional subscript operator that allows assignment
+to an element of the buffer. It behaves as if it returns a reference to that
+element as stored in device memory.
+
+\begin{HLSL}
+device T &operator[](int index);
+\end{HLSL}
+
+\begin{note}
+  Accurately modeling the memory operations for writable typed buffers would
+  require a proxy-object interface since the address written to isn't actually
+  the device memory since it may be converted or otherwise normalized by the
+  implementation.
+\end{note}
+
+\p Writes to typed buffer elements overwrite the entire element, even if only
+part of the data is written. The behavior of partial writes is
+implementation-defined.
+
+\Sec{CheckAccessFullyMapped}{Resources.mapcheck}
+
+\p The mapped status value returned by certain texture and buffer methods can be
+intrepreted by a built-in function:
+
+\begin{HLSL}
+bool CheckAccessFullyMapped(uint status);
+\end{HLSL}
+
+\p This function returns true if the value was accessed from memory that was
+mapped by the runtime implementation. If any part of the return value was from
+unmapped memory, false is returned.
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -9,15 +9,6 @@ and Samplers.
 read-write as denoted by the type of the resource, where the names of read-write
 resources are prefixed with \texttt{RW}.
 
-\begin{note}
-The resource class objects in this clause require language behaviors not
-defined in the current language. Specifically, they utilize constructor
-behaviors, reference types, and concepts (for constraining instantiation). For
-ease of reading this clause uses C++ syntax to represent those language
-constructs. Additionally this clause assumes the introduction of the
-\texttt{device} keyword to denote references to device memory.
-\end{note}
-
 \p Global resource declarations are \textit{bound} to device memory managed by
 an application through a runtime implementation. A resource declaration that is
 not bound is said to be \textit{unbound}. Operations on unbound resources are
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -34,7 +34,7 @@ subscript indexing into the \texttt{ResourceDescriptorHeap}
 
 \Sub{Resource Initialization}{Resources.Special.Initialization}
 
-\p A resource declaration in a global scope are called \textit{global
+\p A resource declaration in a global scope is called a \textit{global
 resource declaration} and represents a binding for external data into a
 program. Global resource declarations are implicitly initialized in an
 implementation-defined way.
@@ -45,7 +45,7 @@ implementation-defined way.
 
 \p The typed buffer class template represents a one-dimensional resource
 containing an array of a single given type. Its contents are indexed by typed
-access to each element. The memory backing a resource need not formatted in the
+access to each element. The memory backing a resource need not be formatted in the
 same type as the resource's element type. If required, typed buffer elements are
 converted upon load in an implementation-defined way.
 
@@ -199,7 +199,7 @@ device T &operator[](int index);
 \end{HLSL}
 
 \begin{itemize}
-  \item \textit{Preconditions:} \texttt{index} \textless the number of elements in the
+  \item \textit{Requires:} \texttt{index} \textless the number of elements in the
   typed buffer.
   \item \textit{Returns:} An lvalue-reference to the element in the typed buffer
   at the index specified by \texttt{index}.
@@ -225,7 +225,7 @@ bool CheckAccessFullyMapped(uint status);
 \end{HLSL}
 
 \begin{itemize}
-  \item \textit{Preconditions:} \texttt{status} is an initialized value from a
+  \item \textit{Requires:} \texttt{status} is an initialized value from a
   resource \texttt{Load} function's \texttt{status} output parameter.
   \item \textit{Returns:} \texttt{true} if the value was accessed from memory
   that was fully mapped by the runtime implementation; otherwise returns
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -90,9 +90,9 @@ class RWBuffer {
 \end{HLSL}
 
 \p The element type of a typed buffer, \texttt{T}, must be an arithmetic type
-(\ref{Basic.Types.Arithmetic}) or vector of arithmetic containing no more than 4
-elements, and the total size of the element (\texttt{sizeof(T)}) must not exceed
-16-bytes in size.
+(\ref{Basic.Types.Arithmetic}) or vector of arithmetic type containing no more
+than 4 elements, and the total size of the element (\texttt{sizeof(T)}) must not
+exceed 16-bytes in size.
 
 \p The \texttt{Buffer} type's template argument \texttt{T} has a default value
 of \texttt{float4}.
--- a/spec/resources.tex
+++ b/spec/resources.tex
@@ -37,6 +37,9 @@ The total size of a single contained element must be less than 16 bytes.
 \p Typed buffers perform format conversions on load such that the underlying
 data gets converted to the destination type in an implementation-defined way.
 
+\p The \texttt{Buffer} type's template argument \texttt{T} has a default value
+of \texttt{float4}.
+
 \begin{HLSL}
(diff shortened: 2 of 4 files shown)


spec/library.tex
```diff
@@ -0,0 +1,31 @@
...
+\Sec{Syntax Notations}{Library.Notation}
+
+\p Some of the constructs in the standard library are not expressible with
+spellable syntax. This section describes syntax constructs which are
+non-standard but are used throughout the standard library specification.
+
+\Sub{Constructors}{Library.Notation.Constructors}
+
+\p Some standard library types have behaviors associated in \gls{isoCPP} with
+constructors. These syntaxes are intended to describe compatible behavior with
```
pow2clk: I'm not sure "compatible" is the right word here. "Comparable" feels too weak. "Consistent" might have the same problem. I landed on "equivalent"
```suggestion
constructors. These syntaxes are intended to describe equivalent behavior with
```

spec/library.tex
```diff
@@ -0,0 +1,31 @@
...
+or return value are references (passed by address or returned by address). To
+simplify this notation the syntax \textit{address-space\textsubscript{opt}}
+\texttt{T \&} is used where the optional \textit{address-space} is one of
+\texttt{device} or \texttt{groupshared}.
+
+\Sub{Concepts}{Library.Notation.Concepts}
+
+\p Some standard library template functions or objects have constraints to which
+types are valid as template arguments. To simplify this notation the concepts
+syntax from \gls{isoCPP20} is used.
```
pow2clk, line 35: This is going to get really verbose in some cases. I fear we are moving in the direction of specifying behavior using code instead of text. I would prefer to state the constraints on parameters. I think that is more consistent with most API documentation

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
+\Ch{Resource Types}{Resources}
+
+\p Resources are intangible types representing memory with an external interface.
```
pow2clk: I think it would be less confusing to refer to the types as "Resource types" and instances of them as "resources"

While I suppose there is some memory associated with a SamplerState object, I wouldn't consider it to represent memory in the same way as a texture or buffer. Maybe something like this?
```suggestion
\p Resource types are intangible types representing memory or configuration with an external interface.
```

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+
+\p These take the form of Typed Buffers, Raw Buffers, Textures, Constant Buffers
+and Samplers.
+
+\p Typed Buffer, Raw Buffer, and Texture resources may be read-only or
+read-write as denoted by the type of the resource, where the names of read-write
+resources are prefixed with \texttt{RW}.
```
pow2clk:
```suggestion
resource types are prefixed with \texttt{RW}.
```

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+an application through a runtime implementation. A resource declaration that is
+not bound is said to be \textit{unbound}. Operations on unbound resources are
+evaluated as no-ops.
+
+\p An implementation may impose restrictions on how local resource objects may
+be used (\ref{AnnexB.LocalResources}).
+
+\Sec{Typed Buffers}{Resources.TypedBuffers}
+
+\p The typed buffer class template represents a one-dimensional resource
```
pow2clk: It's capitalized above
```suggestion
\p The Typed Buffer class template represents a one-dimensional resource
```

spec/resources.tex
pow2clk: Yes, I'm aware I wrote this

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+  // Element access.
+  T operator[](int index) const;
+  device T &operator[](int n) const;
+  T Load(int index) const;
+  T Load(int index, out uint status) const;
+};
+\end{HLSL}
+
+\Sub{Constructors}{Resources.TypedBuffers.Ctrs}
+
```
pow2clk, line 99: I see you removed the language about default template parameters. Removing that support is a breaking change

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+\p This returns the full length, in elements of the buffer through the
+\texttt{width} output parameter.
+
+\Sub{Element Access}{Resources.TypedBuffers.Access}
+
+\p The contents of typed buffers can be retrieved using the \texttt{Load} methods
+or the subscript operator.
+
+\begin{HLSL}
+T Load(in int index) const;
```
pow2clk: As you removed the `in` qualifiers above relying on the reader to assume implicit behavior, it seems consistent to remove it here too
```suggestion
T Load(int index) const;
```

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+
+\p These each return the element of the buffer's type at the given index
+\texttt{index}.
+
+\p An additional \texttt{Load} method returns the indicated element just as the
+other, but also takes a second \texttt{status} parameter that returns
+information about the accessed resource.
+
+\begin{HLSL}
+T Load(in int index, out uint status) const;
```
pow2clk: as above
```suggestion
T Load(int index, out uint status) const;
```

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+\end{HLSL}
+
+\p The \texttt{status} parameter returns an indication of whether all of the
+retrieved value came from fully mapped parts of the resource. This parameter
+must be passed to the built-in function \texttt{CheckAccessFullyMapped}
+(\ref{Resources.Mapcheck}) in order to interpret it.
+
+\p Writable buffers have an additional subscript operator that allows assignment
+to an element of the buffer. It behaves as if it returns a reference to that
+element as stored in device memory.
```
pow2clk: Is this really so different from an mmap concept? I don't think that requires any special specification syntax to indicate that you might be writing into something that isn't actually normal memory. As far as the language is concerned, what difference does it make?

spec/resources.tex
```diff
@@ -0,0 +1,168 @@
...
+\begin{note}
+  Accurately modeling the memory operations for writable typed buffers would
+  require a proxy-object interface since the address written to isn't actually
+  the device memory since it may be converted or otherwise normalized by the
+  implementation.
+\end{note}
+
+\p Writes to typed buffer elements overwrite the entire element, even if only
+part of the data is written. The behavior of partial writes is
+implementation-defined.
```
pow2clk: I think this should be undefined

spec/library.tex
llvm-beanz: Ultimately in the future, we need to move to a world where the constraints are representable in code. I think concepts syntax moves us in that direction. If you have an alternative suggestion please suggest it.

spec/resources.tex
llvm-beanz: I don't think these are proper nouns. They're capitalized in the title, but not the text.

spec/resources.tex
llvm-beanz: The language about default template parameters doesn't belong here. It isn't unique to resource type it applies to _all_ template types and should be part of the templates clause.

spec/resources.tex
llvm-beanz: There is actually a semantic problem that is subtle which I'm trying to address here. The real semantics of the subscript operator for a `RWBuffer<T>` isn't that it actually returns a reference to device memory.

The more correct way to model it would be that the subscript operator returns a proxy object, that when written to generates a call that may convert or otherwise mutate the data before the write occurs. This gets _really_ gnarly when you start factoring in the atomic functions, and I haven't really come up with a better way to spec it than this.

spec/resources.tex
llvm-beanz: What's the argument for undefined rather than implementation-defined? The later lets an implementation do whatever it wants as long as it documents it.

spec/resources.tex
Alexander-Johnston: I think Greg is referring to their capitalisation not in the section title, but the previous section. https://github.com/hlsl-tc57/tc57/pull/65/changes#diff-f43f5ddb995c0acc6ae11cd5c41b19f95a84457e4416d8fe0405b3926e7df1c7R5

llvm-beanz: Ah, I've updated the earlier section.

spec/library.tex
pow2clk: Representing it in code for the purpose of enforcement is a separate matter from documenting it in a specification. I strongly disapprove of using source to represent more than the basic API interface. My alternate suggestion is a list of restrictions after the function prototypes are presented just as has been done successfully for many other specifications

spec/resources.tex
pow2clk: It doesn't even apply to all resource types. Multisampled textures require template arguments: https://godbolt.org/z/qz1MGha58

spec/resources.tex
pow2clk: I am saying that it should be undefined instead of implementation-defined. 

Is there a platform you have in mind here or do you just want to keep options open? Requiring treatment of each element as a unit that requires full updating is pretty close to the definition of a typed buffer. I think failing to do that should be treated as invalid.

spec/resources.tex
llvm-beanz: Whether or not a given resource type has a default argument is explicit in the definition of the type. See the above code:

```
template <typename T = float4> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
class Buffer {
...
};


template <typename T> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
class RWBuffer {
...
};
```

`Buffer` has a default parameter, `RWBuffer` does not. Is there a reason to also explicitly spell this out?

The fact that the default parameter allows the type to be spelled without specifying the type or empty template arguments isn't unique to these templates, that's just HLSL's template behavior.

spec/resources.tex
llvm-beanz: I believe this should be implementation-defined because each implementation of HLSL should specify this behavior, which allows each implementation to have a different behavior as long as it is well-defined.

Why should it be undefined?

spec/library.tex
```diff
@@ -0,0 +1,31 @@
...
+
+\p This clause serves as an overview to the remaining clauses of the
+specification which detail \acrshort{hlsl}'s standard library of functions and
+objects.
+
+\Sec{Syntax Notations}{Library.Notation}
+
+\p Some of the constructs in the standard library are not expressible with
+spellable syntax. This section describes syntax constructs which are
+non-standard but are used throughout the standard library specification.
```
llvm-beanz, line 11:
```suggestion
non-standard but are used throughout the standard library specification.
All code blocks in the library clauses are non-normative.
```

spec/library.tex
llvm-beanz: Do you have a concrete disagreement with this language or the way this PR is proposing text?

In the code snippets in this PR I've used concepts to represent constraints on template instantiation, but also included text describing the constraints.

spec/resources.tex
llvm-beanz: Let me give a concrete example why I believe this _shouldn't_ be undefined behavior.

If we set this as undefined behavior it would be legal for the compiler to optimize the following shader to a no-op:

```hlsl
RWBuffer<float4> Buf;

[numthreads(64,1,1)]
void main(uint3 tid : SV_DispatchThreadID) {
    float4 Val = Buf[tid.x];
    float sum = Val.x + Val.y + Val.z + Val.w;
    Buf[tid.x].x = sum;
}
```
[Compiler Explorer](https://godbolt.org/z/n67s83Gxz)

That shader today compiles to both valid DXIL and valid SPIRV, and it does _something_ in each case. If we decide that a partial write to a typed buffer is undefined, the compiler would legally be allowed to eliminate the write, which would then mean it could legally eliminate the whole rest of the shader body.

That does not seem like the behavior we want.

Instead, if it is implementation defined, the compiler can't eliminate the write it is the responsibility of the implementation to have the partial write _do something_ and to document what that something is.

spec/resources.tex
pow2clk: As a result of the committee's decision that code blocks are non-normative, this isn't going to be sufficient. We'll need some language explaining this as well.

llvm-beanz: Please provide specifics as to what you believe needs to change. The language I removed described how default template arguments behave, which is not something this clause should be addressing. The language [here](https://github.com/hlsl-tc57/tc57/pull/65/changes#diff-f43f5ddb995c0acc6ae11cd5c41b19f95a84457e4416d8fe0405b3926e7df1c7R33) describes the constraints on the template argument, and the (newly added) line [here](https://github.com/hlsl-tc57/tc57/pull/65/changes#diff-f43f5ddb995c0acc6ae11cd5c41b19f95a84457e4416d8fe0405b3926e7df1c7R40) describes the default argument for the `Buffer` type.

Is there any further change required?

spec/library.tex
```diff
@@ -0,0 +1,35 @@
...
+
+\p Code blocks in the remaining clauses of the specification decribe a
+hypothetical implementation and are provided for clarity. A conforming
+implementation need not rely on the specifc code in any of the blocks.
```
llvm-beanz: I've revised the language about code blocks being non-normative. The wording was hastily added in a meeting to try and break an impasse, and I don't believe it actually represents the sentiments expressed by many members of the committee. During the meeting it was expressed that we should represent constraints _both_ in code blocks and in descriptive text, because both are deemed valuable.

We did not want the code blocks to represent an explicit requirement of how the implementations are constructed, and there was some expressed dissatisfaction with technical documentation outside the scope of TC57 that relies heavily on template meta-programming constructs for communicating constraints.

I've put this on the agenda for the next meeting to discuss, but I believe the refined wording more accurately expresses the desire for the specific implementation details to be non-binding, but the code blocks are equally meaningful to the specification as the supporting descriptive text.

https://github.com/hlsl-tc57/tc57/pull/117/changes#diff-57120f4da4883d98975c01e72c6709b1d34c88f87ee106dbbd63efcc50209840R44

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+
+\Sec{Typed Buffers}{Resources.TypedBuffers}
+
+\p The typed buffer class template represents a one-dimensional resource
+containing an array of a single given type. Its contents are indexed by typed
+access to each element. Types may have their formats converted upon load in an
+implementation-defined way.
+
+\p The element type may be any type up to 16 bytes in size. Elements may be
+padded to 16 bytes if the element type is smaller than 16 bytes.
```
Alexander-Johnston: Not that it needs to be done here, but do we need to define padding rules/expected padding? If so, should we add a link to it here? Maybe it's too much detail, but anywhere `may` is used in a spec I have these thoughts.

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+
+\p All typed buffers can be read through subscript operators or Load methods.
+Writable typed buffers can be written through subscript operators.
+
+\p Where \texttt{T} can be any arithmetic type \ref{Basic.Types.Arithmetic}
+or a vector with a maximum of four elements containing such an arithmetic type.
```
Alexander-Johnston: Annex B describes that an implementation may allow vectors of size greater than 4 elements. How does that interact with this clause?

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+The total size of a single contained element must be less than 16 bytes.
+
+\p Typed buffers perform format conversions on load such that the underlying
+data gets converted to the destination type in an implementation-defined way.
+
+\p The \texttt{Buffer} type's template argument \texttt{T} has a default value
+of \texttt{float4}.
+
+\begin{HLSL}
+template <typename T = float4> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
```
Alexander-Johnston: This line is rendering past the edge of the page in the pdf for me. Cuts off half way through `value` at the end.

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+  T Load(int index, out uint status) const;
+};
+
+
+template <typename T> requires hlsl::is_vector<T>::value || hlsl::is_arithmetic<T>::value
+class RWBuffer {
+  RWBuffer();
+  RWBuffer(RWBuffer buf);
+  RWBuffer &operator=(RWBuffer &buf);
+  Buffer(const __DynamicResourceHandle &handle);
```
Alexander-Johnston, line 80: I'm not a fan of defining `__DynamicResourceHandle` here in the code listing then waiting a few points to define it in text. I think the text definition is fine, but I wish it were a little closer to the usage here.

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+\p When defined at local scope, typed buffers represent local references that
+can be associated with global buffers when assigned. Use of an uninitialized
+typed buffer is undefined.
+
+\begin{HLSL}
+Buffer<int> grobuf;
+RWBuffer<int> grwbuf;
+void main() {
+  Buffer<int> lrobuf = grobuf;
+  RWBuffer<int> lrwbuf = grwbuf;
```
Alexander-Johnston: It's not made explicitly clear what `lrobuf` `grobuf` `lrwbuf` and `grwbuf` are here, or if the names hold important meaning. I realise it maps to global/local read write/read only after thinking about it for a moment, but would be nicer if it were a little clearer.

spec/resources.tex
```diff
@@ -0,0 +1,171 @@
...
+
+\p The size of a typed buffer can be retrieved using the \texttt{GetDimensions}
+method.
+
+\begin{HLSL}
+void GetDimensions(out uint width) const;
+\end{HLSL}
+
+\p This returns the full length, in elements of the buffer through the
+\texttt{width} output parameter.
```
Alexander-Johnston: Curiosity rather than issue, but is there a reason we use an out param for `GetDimensions` rather than returning a `uint`. Is this just to keep uniformity with the texture variants of `GetDimensions`?

Alexander-Johnston (commented):
Just a few comments, mostly questions.

To summarise my thoughts a little
* I like that we define the implementation overview in both code and text
* I like that they build from each other
* I don't like ordering, or perhaps the better word is disconnection, between them.

I'm interested in your opinion of defining resources/library code and so on in a structure of
1. Textual overview of the class
2. 1-by-1 definition of class functions (in category order, i.e. constructors, then access, and so on as is already done). These would lay-out a function prototype followed by text describing the purpose/usage then all restrictions.
3. A class listing in code which shows all of the previously defined functions 1-to-1.

At the moment we have a class listing which includes previously undescribed names (I'm looking at `__DynamicResourceHandle` here), then a constructors section which doesn't use/define the code presented in the code listing.
Reordering would help avoid the usage of `__DynamicResourceHandle` far from it's definition.
I think `GetDimensions` is the perfect example of the structure of "define code, explain code, reuse code exactly in class listing" I'm trying to convey here.

spec/library.tex
Alexander-Johnston: I'm personally okay with this wording as it is now. Perhaps it's better to say "Code blocks in sections X Y Z... of the specification" rather than "remaining clauses" just in case there is later shuffling of the specification, but without having a list of sections that's a little hard to do.
Either way, this is releasable text in my mind.

spec/library.tex
Alexander-Johnston: I think the inclusion of restraints in both code and text is useful. So long as both convey exactly the same meaning/there's no ambiguity between them it seems a very useful thing to provide for implementers - 2 separately written texts confirming each other's meaning helps them confirm their understanding, and allows them to read them in the way preferable to them for reference.

spec/resources.tex
llvm-beanz: I agree the padding doesn't belong, and I think the constraint on the element type isn't quite accurate. I'm also noticing that we basically restate this constraint a couple paragraphs down. I'll rework this.

spec/resources.tex
llvm-beanz: I honestly have no idea when this decision was made, the spec is just matching the behavior. I suspect it is historical more than anything.

llvm-beanz:
> I'm interested in your opinion of defining resources/library code and so on in a structure of
> 
> 1. Textual overview of the class
> 2. 1-by-1 definition of class functions (in category order, i.e. constructors, then access, and so on as is already done). These would lay-out a function prototype followed by text describing the purpose/usage then all restrictions.
> 3. A class listing in code which shows all of the previously defined functions 1-to-1.

I'm very open to thoughts on ordering. The current spec follows more or less what the C++ spec does (the `vector` template is maybe a good representative example https://timsong-cpp.github.io/cppwp/n3337/vector). It more or less has a textual overview of the class followed by the class listing, then a structured member-by-member description. I personally really like how C++ does this. In particular I like how the C++ specifications description of methods is much more direct which leaves less room for interpretation.

They have a subclause that describes how they document the library here: https://timsong-cpp.github.io/cppwp/n3337/description.

I think this is something we should do too once we settle on what this format is.
 
> At the moment we have a class listing which includes previously undescribed names (I'm looking at `__DynamicResourceHandle` here), then a constructors section which doesn't use/define the code presented in the code listing. Reordering would help avoid the usage of `__DynamicResourceHandle` far from it's definition.

Yea, I think at a minimum the description of `__DynamicResourceHandle` needs to be higher up, probably at the entry of the clause (since it will apply to other subclauses too).

> I think `GetDimensions` is the perfect example of the structure of "define code, explain code, reuse code exactly in class listing" I'm trying to convey here.

Given that this PR has garnered significant discussion, I'm going to be more opinionated than I initially started. I didn't write this original text it was just coming over from Microsoft's draft. I don't like it.

If we applied C++'s formatting to `GetDimensions` we'd get something more like:

> ```hlsl
> void GetDimensions(out uint width) const;
> ```
> - _Effects:_ Writes the total number of elements contained in the typed buffer to the `width` argument.

This is actually _much_ more concise than what we currently have in this PR, and I think it eliminates a lot of redundant language that isn't actually strictly required.

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
```
bogner:
```suggestion
implementation need not rely on the specific code in any of the blocks.
```

spec/resources.tex
```diff
@@ -0,0 +1,232 @@
+\Ch{Resource Types}{Resources}
+
+\p Resource types are intangible types representing memory or configuration with an external interface.
```
bogner: 80-col
```suggestion
\p Resource types are intangible types representing memory or configuration with
an external interface.
```

spec/resources.tex
```diff
@@ -0,0 +1,232 @@
...
+(\ref{Resources.Binding.Dynamic}).
+
+\Sec{Special Semantics}{Resources.Special}
+
+\Sub{Resource Initialization}{Resources.Special.Initialization}
+
+\p A resource declaration in a global scope are called \textit{global
+resource declaration} and represents a binding for external data into a
+program. Global resource declarations are implicitly initialized in an
+implementation-defined.
```
bogner:
```suggestion
program. Global resource declarations are implicitly initialized in an
implementation-defined way.
```

bogner (approved):
I think the prescriptive approach to defining library types and the additions to the Library Overview make this quite a bit easier to follow, and I think this is much clearer now.

Some grammar nits and typo fixes inline.

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
+
+\Sec{Syntax Notations}{Library.Notation}
+
+\p Some of the constructs in the standard library are not expressible with
+spellable syntax. This section describes syntax constructs which are
+non-standard but are used throughout the standard library specification.
+
+\p Code blocks in the remaining clauses of the specification decribe a
+hypothetical implementation and are provided for clarity. A conforming
+implementation need not rely on the specific code in any of the blocks.
```
pow2clk, line 15: I feel we may have been a bit vague about what the committee agreed to on this matter at the last meeting. I think the original proposal was what was here before, that code blocks are non-normative. The way I read this, it falls a bit short of that. Is that the intent?

We should confirm that this corresponds to how the committee wants code blocks to be viewed.

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
+
+\p The detailed specification for each entity will contain each of the below
+that apply:
+
+\begin{itemize}
+  \item name and description
+  \item class definition or function prototype
+  \item template argument constraints
+  \item class invariants
+  \item function semantics
```
pow2clk: The other lists of sections are capitalized
```suggestion
  \item Name and description
  \item Class definition or function prototype
  \item Template argument constraints
  \item Class invariants
  \item Function semantics
```

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
+
+\begin{itemize}
+  \item \textit{Requires:} the preconditions for calling the function.
+  \item \textit{Effects:} the actions performed by the function.
+  \item \textit{Synchronization:} the requirements for execution or memory
+  synchronization.
+  \item \textit{Returns:} a description of the value returned by the function.
+  \item \textit{Remarks:} additional normative information about the function.
+  \item \textit{Notes:} non-normative information provided about the function.
+\end{itemize}
```
pow2clk, line 112: I can't see anything wrong with this as a suggested template except that it is so dogmatic about the specific order and potential sections. There may be sections that would be better expressed with something that deviates a bit from this. I don't have examples, but it seems a bit peculiar to have a portion of the spec that dictates how the spec should be written.

spec/resources.tex
```diff
@@ -71,98 +89,145 @@
...
-part of the data is written. The behavior of partial writes is
-implementation-defined.
-
-\Sec{CheckAccessFullyMapped}{Resources.mapcheck}
+\Sec{CheckAccessFullyMapped}{Resources.Mapcheck}
 
-\p The mapped status value returned by certain texture and buffer methods can be
-intrepreted by a built-in function:
+\p The implementation-defined status value returned by texture and buffer load
+methods is interpreted by a built-in function:
```
pow2clk, line 221: Why was the description of the status parameter altered only here to merely be identified as "implementation-defined"?

spec/resources.tex
```diff
@@ -71,98 +89,145 @@
...
 \end{HLSL}
 
+\begin{itemize}
+  \item \textit{Preconditions:} \texttt{index} \textless the number of elements in the
+  typed buffer.
+  \item \textit{Returns:} An lvalue-reference to the element in the typed buffer
+  at the index specified by \texttt{index}.
+  \item \textit{Remarks:} Writing to typed buffer elements overwrite the entire
+  element, even if only part of the data is written. A partial write to a typed
+  buffer is a load-insert-store operation.
```
pow2clk, line 208: Why the change from "implementation-defined"? Where is "load-insert-store" described?

spec/resources.tex
```diff
@@ -17,28 +18,44 @@
...
-access to each element. Types may have their formats converted upon load in an
+\p \textit{Dynamic resources} is an optional feature which allows resource
+objects to be initialized by indexing directly into a heap of descriptors. A
+conforming implementation must either support all interfaces related to dynamic
+resources or none, but may not partially support them.
+
+\p The \texttt{\_\_DynamicResourceHandle} type in the code examples in this
+clause is a placeholder for the implementation-specific type returned by
+subscript indexing into the \texttt{ResourceDescriptorHeap}
+(\ref{Resources.Binding.Dynamic}).
```
pow2clk, line 31: This section doesn't exist and was missing from hlsl-specs as well.

spec/resources.tex
```diff
@@ -17,28 +18,44 @@
...
+
+\Sec{Special Semantics}{Resources.Special}
+
+\Sub{Resource Initialization}{Resources.Special.Initialization}
+
+\p A resource declaration in a global scope are called \textit{global
+resource declaration} and represents a binding for external data into a
+program. Global resource declarations are implicitly initialized in an
```
pow2clk, line 39: Is the intent to add the Resources.Binding section eventually or just leave it at "implementation-defined way"?

spec/resources.tex
```diff
@@ -17,28 +18,44 @@
...
-padded to 16 bytes if the element type is smaller than 16 bytes.
+\p Use of an uninitialized resource is undefined.
 
-\p All typed buffers can be read through subscript operators or Load methods.
-Writable typed buffers can be written through subscript operators.
+\Sec{Typed Buffers}{Resources.TypedBuffers}
 
-\p Where \texttt{T} can be any arithmetic type \ref{Basic.Types.Arithmetic}
-or a vector with a maximum of four elements containing such an arithmetic type.
-The total size of a single contained element must be less than 16 bytes.
```
pow2clk: Why remove this information? At the least, mention that a type exceeding a certain size or number of elements may be disallowed by an implementation

spec/resources.tex
```diff
@@ -17,28 +18,44 @@
...
+objects to be initialized by indexing directly into a heap of descriptors. A
+conforming implementation must either support all interfaces related to dynamic
+resources or none, but may not partially support them.
+
+\p The \texttt{\_\_DynamicResourceHandle} type in the code examples in this
+clause is a placeholder for the implementation-specific type returned by
+subscript indexing into the \texttt{ResourceDescriptorHeap}
+(\ref{Resources.Binding.Dynamic}).
+
+\Sec{Special Semantics}{Resources.Special}
```
pow2clk, line 33: Oh I don't like that the subsection has to be called this. It doesn't reveal much of what the content is.

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+
+\Sub{Resource Initialization}{Resources.Special.Initialization}
+
+\p A resource declaration in a global scope are called \textit{global
```
Alexander-Johnston:
```suggestion
\p A resource declaration in a global scope is called a \textit{global
```

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+program. Global resource declarations are implicitly initialized in an
+implementation-defined way.
+
+\p Use of an uninitialized resource is undefined.
+
+\Sec{Typed Buffers}{Resources.TypedBuffers}
+
+\p The typed buffer class template represents a one-dimensional resource
+containing an array of a single given type. Its contents are indexed by typed
+access to each element. The memory backing a resource need not formatted in the
```
Alexander-Johnston:
```suggestion
access to each element. The memory backing a resource need not be formatted in the
```

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+\end{itemize}
+
+\Sub{Typed Buffer Operators}{Resources.TypedBuffers.Operators}
+
+\begin{HLSL}
+T operator[](uint index) const;
+\end{HLSL}
+
+\begin{itemize}
+  \item \textit{Requires:} \texttt{index} \textless the number of elements in the
```
Alexander-Johnston: I know you pushed a commit to fix the rendering from `<` to `\textless`, but I fear it's still rending incorrectly. I'm not seeing a space between  `<` and `the`. Same with the future uses of `\textless`. Very very minor.

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+  \item \textit{Returns:} The element in the typed buffer at the index specified
+  by \texttt{index}.
+\end{itemize}
+
+\begin{HLSL}
+device T &operator[](int index);
+\end{HLSL}
+
+\begin{itemize}
+  \item \textit{Preconditions:} \texttt{index} \textless the number of elements in the
```
Alexander-Johnston: For the previous operator you used `Requires` but here `Preconditions` for what is the same requirement. I believe both should be `Requires`

```suggestion
  \item \textit{Requires:} \texttt{index} \textless the number of elements in the
```

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+
+\p The implementation-defined status value returned by texture and buffer load
+methods is interpreted by a built-in function:
+
+\begin{HLSL}
+bool CheckAccessFullyMapped(uint status);
+\end{HLSL}
+
+\begin{itemize}
+  \item \textit{Preconditions:} \texttt{status} is an initialized value from a
```
Alexander-Johnston:
```suggestion
  \item \textit{Requires:} \texttt{status} is an initialized value from a
```
Same as above.

Alexander-Johnston (approved):
I think this reads much better now. I appreciate the opinionated rewrite!
I don't know that this is the _perfect_ way to structure the definition of all library material, but I like that it prescribes a method that should be consistent throughout the specification, making it much easier to skim through and use as reference.

I don't have any issues with the current wording intention, just a few typos and such along with https://github.com/hlsl-tc57/tc57/pull/65/changes/3ef2a6202663ffbf24ff97238aab69171034c729#r3532373239

spec/library.tex
llvm-beanz: I left a comment about this [here](https://github.com/hlsl-tc57/tc57/pull/65/#discussion_r3467247576) in which I said:

> The wording was hastily added in a meeting to try and break an impasse, and I don't believe it actually represents the sentiments expressed by many members of the committee.

I don't believe from the discussion in the meeting that we actually want code blocks to be non-normative, we want them to be supplemented with descriptive text.

@Alexander-Johnston's comment [here](https://github.com/hlsl-tc57/tc57/pull/65/#discussion_r3491821045) seems to concur with my assessment, as does @bogner's approval.

spec/library.tex
llvm-beanz: As @bogner says [here](https://github.com/hlsl-tc57/tc57/pull/65/#pullrequestreview-4603409979)

> I think the prescriptive approach to defining library types and the additions to the Library Overview make this quite a bit easier to follow, and I think this is much clearer now.

As @Alexander-Johnston says [here](https://github.com/hlsl-tc57/tc57/pull/65/#pullrequestreview-4640591579):

> I don't know that this is the perfect way to structure the definition of all library material, but I like that it prescribes a method that should be consistent throughout the specification, making it much easier to skim through and use as reference.

I believe we have two reviews confirming this approach and one dissenting view, so unless there are further dissents I believe the current approach is correct.

If you have suggestions on how to make the prescriptive template more palatable to you I'm open to them, but I believe the current wording aligns with the overall preference as expressed thus far.

spec/resources.tex
llvm-beanz: In many other PRs I've created references to sections that don't yet exist. LaTeX provides nice diagnostics of these that are easy to track. If this PR is approved by the committee I'll file an issue for the unwritten section to track populating it.

spec/resources.tex
llvm-beanz: Do you have an alternative suggestion?

spec/resources.tex
llvm-beanz: I expect to add a section that describes the resource binding attributes, but not the specific semantics of binding since they vary by implementation, so this statement will remain valid.

spec/resources.tex
llvm-beanz: @bogner and I investigated and found that DXC and Clang for both DirectX and Vulkan behave consistently so we decided to codify that behavior.

spec/resources.tex
llvm-beanz: It is identified [here](https://github.com/hlsl-tc57/tc57/pull/65/changes#diff-f43f5ddb995c0acc6ae11cd5c41b19f95a84457e4416d8fe0405b3926e7df1c7R176) also as being implementation-defined. Are there other places that you think it should be clarified that its value is implementation defined?

spec/resources.tex
llvm-beanz: I'm going to mark this conversation as resolved, because this language has been revised. As I commented [here](https://github.com/hlsl-tc57/tc57/pull/65/#discussion_r3533160073), @bogner and I investigated and found all implementations agree on this behavior so it can be specified and defined rather than implantation defined or undefined.

spec/resources.tex
llvm-beanz: Given the extent of the churn on this, it is probably easier to review the final state of the diff rather than individual commits. This information is still present in the specification, it just isn't at this location in the file. It is [here](https://github.com/hlsl-tc57/tc57/pull/65/changes#diff-f43f5ddb995c0acc6ae11cd5c41b19f95a84457e4416d8fe0405b3926e7df1c7R95)

llvm-beanz:
We only briefly discussed this PR in the 2026-07-07 meeting enough to identify that @pow2clk still has concerns and is withholding consensus.

I have no planned changes to the PR at this point waiting on specific feedback. Please provide feedback promptly so that we can move this forward before our next meeting.

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+implementation-defined way.
+
+\p Use of an uninitialized resource is undefined.
+
+\Sec{Typed Buffers}{Resources.TypedBuffers}
+
+\p The typed buffer class template represents a one-dimensional resource
+containing an array of a single given type. Its contents are indexed by typed
+access to each element. The memory backing a resource need not be formatted in the
+same type as the resource's element type. If required, typed buffer elements are
```
jessebarker, line 49: This might just be my own mental model, but "formatted" seems odd here.  I'm accustomed to talking about memory layout in this kind of context ("The layout of the memory backing a resource need not match the layout of the resource's element type.").  If I'm alone here (or if I'm wrong or missed the point :-) ), I definitely don't feel strongly.

llvm-beanz: The wording is definitely odd. The tough thing is that "layout" doesn't quite capture the nuance. The shader may specify that a typed buffer is comprised of 4 16-bit floating point values, but the underlying data could be 4 32-bit integers which are interpolated into 16-bit float values, so it can have both layout and type conversions (with interpolation rather than traditional conversions).

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+  // Element access.
+  T operator[](int index) const;
+  device T &operator[](int n) const;
+  T Load(int index) const;
+  T Load(int index, out uint status) const;
+};
+\end{HLSL}
+
+\p The element type of a typed buffer, \texttt{T}, must be an arithmetic type
+(\ref{Basic.Types.Arithmetic}) or vector of arithmetic containing no more than 4
```
jessebarker: "...vector of arithmetic type..."

spec/resources.tex
```diff
@@ -0,0 +1,233 @@
...
+elements, and the total size of the element (\texttt{sizeof(T)}) must not exceed
+16-bytes in size.
```
jessebarker: 16-bytes seems an odd limit.  Is this not meant to cover structured buffers?

llvm-beanz: Yea, this only covers "typed buffers" which are the `Buffer` and `RWBuffer` objects. Structured buffers are "raw" buffers which have different rules.

spec/resources.tex
jessebarker: That's fair.  We can always update it if we find something more precise.

spec/resources.tex
jessebarker: Cool, thanks.

jessebarker (approved):
Thanks for your answers.  LGTM

spec/library.tex
llvm-beanz: As a note here: this did come up very briefly in the TC57 meeting last week, but we didn't fully discuss it. We will be asking for an affirmation from the delegates at the 2026-07-21 meeting to confirm that this language and the remaining text in the PR matches the committee's intent from 2026-06-23.

cc @pow2clk

llvm-beanz:
@pow2clk still waiting on your feedback.

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
+  \item Inspection functions
+  \item Operators
+  \item Other non-member functions
+\end{itemize}
+
+\p Descriptions of function semantics will contain all of the following elements
+that apply in the order as specified:
+
+\begin{itemize}
+  \item \textit{Requires:} the preconditions for calling the function.
```
pow2clk, line 105: Perhaps this is meant to cover parameter requirements, but the word "preconditions" makes it sound like already configured state rather than how the function is called. I'd feel more comfortable if there were an item for parameter requirements. 
```suggestion
  \item \textit{Requires:} the preconditions for calling the function.
  \item \textit{Parameters:} the requirements of the parameter types and values.
```

spec/library.tex
```diff
@@ -0,0 +1,112 @@
...
+
+\Sub{Clause Structure}{Libarry.Method.Structure}
+
+\p Each library clause will contain all applicable subclauses of the list:
+
+\begin{itemize}
+  \item Introduction
+  \item Optional features
+  \item Special semantics
+  \item Element subclauses
```
pow2clk, line 47: I still feel that these are too strict in particular that the sections need bear these names as they aren't very descriptive, but I'm willing to accept that this section will be updated.

pow2clk (commented):
I didn't have much more to say after our conversation earlier this week than I thought, but ultimately decided to make two small suggestions. My principal concern continues to be that the parameter requirements are represented in the text description, while I'm uncomfortable with some aspects of the library clause template, it can provide that assurance.

spec/library.tex
llvm-beanz: The C++ specification uses a single heading to cover all kinds of requirements (parameters or general program state). I'm not opposed to having different labels, but I'm unsure I see how that increases readability. A few examples of this template applied in C++ that may be helpful to understand how it would apply for our specification:

* [string.accessors] uses _Requires_ to denote program requirements: https://timsong-cpp.github.io/cppwp/n3337/string.accessors#3
* [util.smartptr.shared.atomic] lists multiple _Requires_ annotations under a single API: https://timsong-cpp.github.io/cppwp/n3337/util.smartptr.shared.atomic#17

spec/library.tex
llvm-beanz: Do you have a requested or suggested change here?

spec/library.tex
pow2clk: I think making it explicit makes it more readable and ensures that those trying to follow this section's requirements will be sure to consider parameter requirements

spec/library.tex
pow2clk: I do not. "I'm willing to accept that this section will be updated" was my attempt to say that I'm willing to accept this. It is merely informative.

spec/library.tex
pow2clk: That said, on the principal that something shouldn't be blocked for something it lacks, I'll create an issue for this and we can address it at another time.

llvm-beanz:
Consensus to merge this PR was reached at the 2026-07-21 meeting.

PR #68, opened 2026-06-08
[New] Add proposal for Loop unroll factor

farzonl:
This change brings over the accepted HLSL loop unroll propsal from HLSL specs. The change is essentially to treat loop unroll the same way that every other C like language does which is as a hint to the compiler to do partial to full loop unroll. After this change DXC's behavior of specifying a maximum number of loop executions will not be carried forward.

### DIFF
--- a/proposals/0012-hlsl-loop-unroll-factor.md
+++ b/proposals/0012-hlsl-loop-unroll-factor.md
@@ -1,5 +1,5 @@
 ---
-title: "XXXX - HLSL Loop Unroll Factor"
+title: "0012 - HLSL Loop Unroll Factor"
 params:
   authors:
     - farzonl: Farzon Lotfi
--- a/proposals/XXXX-hlsl-loop-unroll-factor.md
+++ b/proposals/XXXX-hlsl-loop-unroll-factor.md
@@ -19,7 +19,7 @@ This proposal seeks to unify the diverging behaviors between the agreed upon
 and the current behavior of DXC.
 
 The new spec wants to treat the unroll factor as a hint to the compiler for
-partial loop unroll. The current behavior in dxc is that the unroll factor
+partial loop unroll. The current behavior in DXC is that the unroll factor
 specifies the maximum number of times the loop is to execute.
 
 The DXC behavior was determined to violate user expectations by overriding
@@ -33,8 +33,8 @@ implementations that exist in LLVM.
 The HLSL compilers transition to clang has resulted in a compat break between
 the [HLSL loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
 and the one in DXC. While there is an expectation that the new compiler will
-not be fully compatible with the previous compilers, These compat breaks should
- be minimized when possible.
+not be fully compatible with the previous compilers, these compatibility  breaks
+should be minimized when possible. This is not one of those cases.
 
 ## Proposed solution
 In [HLSL 202X](0001-hlsl-202x-202y.md) the DXC implementation of the loop
--- a/proposals/XXXX-hlsl-loop-unroll-factor.md
+++ b/proposals/XXXX-hlsl-loop-unroll-factor.md
@@ -1,5 +1,5 @@
 ---
-title: "0007 - HLSL Loop Unroll Factor"
+title: "XXXX - HLSL Loop Unroll Factor"
 params:
   authors:
     - farzonl: Farzon Lotfi
--- a/proposals/0007-hlsl-loop-unroll-factor.md
+++ b/proposals/0007-hlsl-loop-unroll-factor.md
@@ -1,0 +1,49 @@
+---
+title: "0007 - HLSL Loop Unroll Factor"
+params:
+  authors:
+    - farzonl: Farzon Lotfi
+  sponsors:
+    - farzonl: Farzon Lotfi
+  status: Refinement
+---
+
+* Planned Version: 202x
+* PRs: [#276](https://github.com/microsoft/hlsl-specs/pull/276)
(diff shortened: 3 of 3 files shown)


proposals/XXXX-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,49 @@
...
+made our port of HLSL loop unroll a bunch special cases instead of just
+syntactic sugar that could sit on top of the existing loop unroll
+implementations that exist in LLVM.
+
+## Motivation
+The HLSL compilers transition to clang has resulted in a compat break between
+the [HLSL loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
+and the one in DXC. While there is an expectation that the new compiler will
+not be fully compatible with the previous compilers, These compat breaks should
+ be minimized when possible.
```
pow2clk: typos
```suggestion
not be fully compatible with the previous compilers, these compatibility breaks should
be minimized when possible.
```

proposals/XXXX-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,49 @@
...
+In [HLSL 202X](0001-hlsl-202x-202y.md) the DXC implementation of the loop
+unroll factor should no longer alter the  number of iteration executions.
+Ideally the loop unroll factor will also become a compiler hint for partial
+loop unroll. If this can not be done because of concerns of invalid DXIL
+generation then the loop unroll factor as a feature should be removed from DXC.
+Removal in this case means ignoring the unroll factor and issuing a warning to
+inform the user.
+
+## Detailed design
+Full specification language was accepted in [#263](https://github.com/microsoft/hlsl-specs/pull/263)
```
pow2clk: Is this section still TBD or is the spec language the detailed expected behavior? If it is, what is this except fixing a bug in DXC where it doesn't reflect that language?

proposals/XXXX-hlsl-loop-unroll-factor.md
pow2clk: Are we minimizing compatibility breaks here by changing the behavior of DXC? That doesn't sound so much like minimizing the breaks as introducing them early

proposals/XXXX-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,49 @@
...
+* Impacted Projects: DXC & Clang
+
+## Introduction
+
+This proposal seeks to unify the diverging behaviors between the agreed upon
+[spec for HLSL loop unroll](https://github.com/microsoft/hlsl-specs/pull/263)
+and the current behavior of DXC.
+
+The new spec wants to treat the unroll factor as a hint to the compiler for
+partial loop unroll. The current behavior in dxc is that the unroll factor
```
pow2clk: I don't really care if we capitalize or not, but we should be consistent
```suggestion
partial loop unroll. The current behavior in DXC is that the unroll factor
```

proposals/0012-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Introduction
+
+This proposal seeks to unify the diverging behaviors between the agreed upon
+[spec for HLSL loop unroll](https://github.com/microsoft/hlsl-specs/pull/263)
+and the current behavior of DXC.
+
+The new spec wants to treat the unroll factor as a hint to the compiler for
+partial loop unroll. The current behavior in dxc is that the unroll factor
+specifies the maximum number of times the loop is to execute.
```
pow2clk, line 23: I think it's worth mentioning that it only does this when it determines that unrolling is possible

pow2clk (commented):
While I fully agree that the DXC behavior should change, I don't think this is a language proposal. I think DXC has a bug

farzonl:
> While I fully agree that the DXC behavior should change, I don't think this is a language proposal. I think DXC has a bug


The documented behavior is:
> specify the maximum number of times the loop is to execute. 

Short of an existing spec we have to go with what the docs say which is also the current behavior of DXC. So to me it looks like DXC faithfully executed what we said it should have been. At this point it is a chicken and egg problem to say whether the documentation or the DXC behavior came first so if this is a bug we still need a language proposal to tell the community of this breaking change going forward.

 https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-for

pow2clk:
> https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-for

Yeah, I see that now. I find it surprising, but that's enough to convince me it is working as intended. I agree it should be changed.

proposals/XXXX-hlsl-loop-unroll-factor.md
farzonl: This section is not tbd we adopted new language in the LaTex to match closer to Clang and OpenCLs loop unroll hints do.

proposals/XXXX-hlsl-loop-unroll-factor.md
farzonl: Will clarify that this is not one of the minimization cases.

proposals/0012-hlsl-loop-unroll-factor.md
farzonl: Unroll is a hint in both compilers that;s not really a unique feature between the two so don't know why we want to call it out. just sounds verbose.

pow2clk (approved):

llvm-beanz:

PR #72, opened 2026-06-18
[New][0013] Add proposal for named casts

llvm-beanz:
This proposal adds new named casts for HLSL.

### DIFF
--- a/proposals/0013-named-casts.md
+++ b/proposals/0013-named-casts.md
@@ -1,5 +1,5 @@
 ---
-title: "NNNN - Named Casts"
+title: "0013 - Named Casts"
 draft: true
 params:
   authors:
--- a/proposals/NNNN-named-casts.md
+++ b/proposals/NNNN-named-casts.md
@@ -26,7 +26,7 @@ values from a source type to a destination type with tight validation.
 ## Proposed solution
 
 HLSL will introduce three new named cast templates, `static_cast<T>`,
-`elementwise_cast<T>` and `bit_cast<T>`.
+`elementwise_cast<T>` and `hlsl::bit_cast<T>`.
 
 The `static_cast<T>(V)` will behave similar to the C++ `static_cast<T>(V)`
 expression, except as changes are required due to other differences between C++
@@ -39,7 +39,10 @@ The `elementwise_cast<T>(V)`, will perform an element-wise conversion
 (\ref{Conv.Flat}). The type of the expression `V` and the result type `T` must
 be scalar layout compatible (\ref{Basic.Types.Scalarized}).
 
-The `bit_cast<T>(V)` cast operator will convert the expression `V` to type `T`
-without modifying the precise bit representation. The type of the expression `V`
-and the result type `T` must not be intangible types
+The `hlsl::bit_cast<T>(V)` cast function template will convert the expression
+`V` to type `T` without modifying the precise bit representation. The type of
+the expression `V` and the result type `T` must not be intangible types
 (\ref{Basic.Types.Intangible}), and must be of the same size.
+
+> Note: the `bit_cast` template is put in the hlsl namespace since it will be a
+> function template matching how C++ implemented `std::bit_cast`.
--- a/proposals/NNNN-named-casts.md
+++ b/proposals/NNNN-named-casts.md
@@ -26,7 +26,7 @@ values from a source type to a destination type with tight validation.
 ## Proposed solution
 
 HLSL will introduce three new named cast templates, `static_cast<T>`,
-`elementwise_cast<T>` and `hlsl::bit_cast<T>`.
+`hlsl::elementwise_cast<T>` and `hlsl::bit_cast<T>`.
 
 The `static_cast<T>(V)` will behave similar to the C++ `static_cast<T>(V)`
 expression, except as changes are required due to other differences between C++
@@ -35,7 +35,7 @@ difference from C++ is that HLSL's `static_cast<T>` will always produce an
 rvalue since HLSL does not have spellable lvalue reference types, and may not be
 used for polymorphic casts for the same reason.
 
-The `elementwise_cast<T>(V)`, will perform an element-wise conversion
+The `hlsl::elementwise_cast<T>(V)`, will perform an element-wise conversion
 (\ref{Conv.Flat}). The type of the expression `V` and the result type `T` must
 be scalar layout compatible (\ref{Basic.Types.Scalarized}).
 
@@ -44,5 +44,7 @@ The `hlsl::bit_cast<T>(V)` cast function template will convert the expression
 the expression `V` and the result type `T` must not be intangible types
 (\ref{Basic.Types.Intangible}), and must be of the same size.
 
-> Note: the `bit_cast` template is put in the hlsl namespace since it will be a
-> function template matching how C++ implemented `std::bit_cast`.
+The `bit_cast` and `elementwise_cast` templates areput in the hlsl namespace
+aligning with modern C++ conventions to avoid additions to the global namespace.
+Conversely the `static_cast` construct is defined as a global casting operator
+conforming to the historical C++ definition which remains in modern C++.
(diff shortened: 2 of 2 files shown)


proposals/NNNN-named-casts.md
```diff
@@ -0,0 +1,48 @@
...
+(\ref{Conv.Flat}). The type of the expression `V` and the result type `T` must
+be scalar layout compatible (\ref{Basic.Types.Scalarized}).
+
+The `hlsl::bit_cast<T>(V)` cast function template will convert the expression
+`V` to type `T` without modifying the precise bit representation. The type of
+the expression `V` and the result type `T` must not be intangible types
+(\ref{Basic.Types.Intangible}), and must be of the same size.
+
+> Note: the `bit_cast` template is put in the hlsl namespace since it will be a
+> function template matching how C++ implemented `std::bit_cast`.
```
Alexander-Johnston: For the sake of consistency I'm not a fan of having some casts inside the hlsl namespace and some outside. (I realise C++ does this but I'm unfamiliar with the history of std::bit_cast that led to it).
Given that we have the namespace to help avoid conflict with C++ is it not reasonable to also put static_cast inside hlsl::?

proposals/NNNN-named-casts.md
```diff
@@ -0,0 +1,48 @@
...
+transformation a cast performs. These take the forms `static_cast`,
+`reinterpret_cast`, `const_cast`, `dynamic_cast`, and most recently `bit_cast`.
+
+This proposal introduces named casts for HLSL to allow unambiguous ways to cast
+values from a source type to a destination type with tight validation.
+
+## Proposed solution
+
+HLSL will introduce three new named cast templates, `static_cast<T>`,
+`elementwise_cast<T>` and `hlsl::bit_cast<T>`.
```
Alexander-Johnston: Why not include a `const_cast`? Is it because the intention is to allow other casts to cast away the const?
Or because we don't want a const_cast for explicit reasons? If the latter, we should specify that casting away const is invalid.
Though perhaps I'm overthinking, as I suppose it doesn't matter too much either way if we're always producing rvalues from the casts and we aren't supporting references or pointers.

proposals/0013-named-casts.md
```diff
@@ -0,0 +1,48 @@
...
+
+HLSL will introduce three new named cast templates, `static_cast<T>`,
+`elementwise_cast<T>` and `hlsl::bit_cast<T>`.
+
+The `static_cast<T>(V)` will behave similar to the C++ `static_cast<T>(V)`
+expression, except as changes are required due to other differences between C++
+and HLSL. It will convert the expression `V` to the result type `T`. The main
+difference from C++ is that HLSL's `static_cast<T>` will always produce an
+rvalue since HLSL does not have spellable lvalue reference types, and may not be
+used for polymorphic casts for the same reason.
```
Alexander-Johnston, line 36: You say the "main difference", but are there other minor differences that should be made clear here?

proposals/NNNN-named-casts.md
Alexander-Johnston: Oh ignore this, I started the review earlier in the day and didn't realise you'd added more commits since.

proposals/NNNN-named-casts.md
llvm-beanz: Your last point is spot on. In C++ `const_cast` only operates on pointers and references. Since HLSL has neither anything you could do with a `const_cast` you could do by creating a copy.

proposals/NNNN-named-casts.md
```diff
@@ -0,0 +1,50 @@
...
+The `hlsl::elementwise_cast<T>(V)`, will perform an element-wise conversion
+(\ref{Conv.Flat}). The type of the expression `V` and the result type `T` must
+be scalar layout compatible (\ref{Basic.Types.Scalarized}).
+
+The `hlsl::bit_cast<T>(V)` cast function template will convert the expression
+`V` to type `T` without modifying the precise bit representation. The type of
+the expression `V` and the result type `T` must not be intangible types
+(\ref{Basic.Types.Intangible}), and must be of the same size.
+
+The `bit_cast` and `elementwise_cast` templates areput in the hlsl namespace
```
Alexander-Johnston:
```suggestion
The `bit_cast` and `elementwise_cast` templates are put in the hlsl namespace
```

proposals/0013-named-casts.md
llvm-beanz: There's some other casts related to pointers that the C++ static cast can be used for which don't apply to HLSL. The full C++ language is available here:

https://timsong-cpp.github.io/cppwp/n3337/expr.static.cast

I expect that we'd have a trimmed down set of language for this.

proposals/NNNN-named-casts.md
Alexander-Johnston: Okay, glad to have my heatwave induced ravings confirmed, thanks!

proposals/NNNN-named-casts.md
llvm-beanz: I had a conversation with @bogner about this, and he suggested we add everything that isn't part of C++'s global namespace to the `hlsl` namespace. That made sense to me.

proposals/0013-named-casts.md
Alexander-Johnston: I suppose this question of mine is just the result of leaning on C++. I like explicitness, but that's for the spec. Thanks for the patience!

Alexander-Johnston (approved):

proposals/NNNN-named-casts.md
```diff
@@ -0,0 +1,50 @@
...
+`hlsl::elementwise_cast<T>` and `hlsl::bit_cast<T>`.
+
+The `static_cast<T>(V)` will behave similar to the C++ `static_cast<T>(V)`
+expression, except as changes are required due to other differences between C++
+and HLSL. It will convert the expression `V` to the result type `T`. The main
+difference from C++ is that HLSL's `static_cast<T>` will always produce an
+rvalue since HLSL does not have spellable lvalue reference types, and may not be
+used for polymorphic casts for the same reason.
+
+The `hlsl::elementwise_cast<T>(V)`, will perform an element-wise conversion
```
pow2clk:
```suggestion
The `hlsl::elementwise_cast<T>(V)` will perform an element-wise conversion
```

proposals/0013-named-casts.md
```diff
@@ -0,0 +1,50 @@
...
+
+The `hlsl::bit_cast<T>(V)` cast function template will convert the expression
+`V` to type `T` without modifying the precise bit representation. The type of
+the expression `V` and the result type `T` must not be intangible types
+(\ref{Basic.Types.Intangible}), and must be of the same size.
+
+The `bit_cast` and `elementwise_cast` templates are put in the hlsl namespace
+aligning with modern C++ conventions to avoid additions to the global namespace.
+Conversely the `static_cast` construct is defined as a global casting operator
+conforming to the historical C++ definition which remains in modern C++.
```
pow2clk, line 50: This is so that c++ and HLSL can share code better? Otherwise, I see no reason to repeat the mistakes of C++ past. Even as it is, I'm not sure the benefit outweighs the disadvantage.

llvm-beanz: Yes, this is to make it easier to share code between C++ and HLSL.

bogner (approved):
LGTM. The divide between namespaces to match C++'s historical artifacts is in some ways awkward but I think it will be best for users nonetheless. 

I would be open to an argument that `static_cast` *also* be available in the `hlsl::` namespace:
- Pro: these could be used consistently in code that is not meant to be shared with C++
- Con: there are two ways to do things

pow2clk:
> LGTM. The divide between namespaces to match C++'s historical artifacts is in some ways awkward but I think it will be best for users nonetheless.
> 
> I would be open to an argument that `static_cast` _also_ be available in the `hlsl::` namespace:
> 
> * Pro: these could be used consistently in code that is not meant to be shared with C++
> * Con: there are two ways to do things

I support including `static_cast` in the `hlsl::` namespace

llvm-beanz:

PR #73, opened 2026-06-19
[New] Add groupshared arguments proposal

llvm-beanz:
This proposal has already been implemented in both DXC and Clang for both SPIR-V and DXIL, and is essential to moving many standard library APIs into spellable syntax.

### DIFF
--- a/proposals/0014-groupshared-arguments.md
+++ b/proposals/0014-groupshared-arguments.md
@@ -1,5 +1,5 @@
 ---
-title: "NNNN - `groupshared` Arguments"
+title: "0014 - `groupshared` Arguments"
 params:
   authors:
     - llvm-beanz: Chris Bieneman
--- a/proposals/NNNN-groupshared-arguments.md
+++ b/proposals/NNNN-groupshared-arguments.md
@@ -39,6 +39,7 @@ parameter declarations. When applied to a parameter declaration of
 type `T`, the keyword alters the qualified type of the parameter to a `groupshared T &`
 (a reference to `groupshared` memory of type `T`).
 
+```
 void fn(groupshared uint4 A) {}
 ```
 
@@ -67,6 +68,7 @@ void fn(groupshared uint4 A) {
 ```
 
 Not Allowed:
+```
 void fn(groupshared uint4 A) {}
 void fn2() {
   float4 B = 1.0.xxxx;
@@ -93,19 +95,10 @@ be finalized.
 
 ### Overloading behavior
 
-void fn(groupshared uint shared);
-void fn(inout uint u);
+Overload sets including value types and `groupshared` arguments of the same
+underlying type are ambiguous if the call site argument is `groupshared`.
 
-groupshared uint Shared;
-
-void caller() {
-  fn(Shared); // ambiguous
-
-  uint Local;
-  fn(Local); // Not ambiguous
-}
 ```
-
 void fn(groupshared uint shared);
 void fn(uint u);
 
@@ -119,11 +112,6 @@ void caller() {
   fn(5); // Not ambiguous
 }
 ```
-
-The above overload sets will result in an error that the call is ambiguous
-when the call site argument is a `groupshared` variable.  They will not result
-in an error at the call site if the argument is a local varible or a literal
-value.
 
 ### Limitations
 
--- a/proposals/NNNN-groupshared-arguments.md
+++ b/proposals/NNNN-groupshared-arguments.md
@@ -42,17 +42,26 @@ type `T`, the keyword alters the qualified type of the parameter to a `groupshar
 void fn(groupshared uint4 A) {}
 ```
 
-No implicit or explicit conversion can change the memory space of an object. To
-perform such a conversion, a user must declare a new object in the destination
-memory space and initialize it appropriately. For overload resolution, the
-parameter type must be an exact match in order for overload resolution to
-succeed since no conversions will be valid.
+No implicit or explicit conversion can change the memory space of an object.
+LValue-to-RValue conversions load a value out of its specified address space
+into a temporary value in the default address space. This is an implicit copy,
+not an address space conversion. To perform such a conversion, a user must
+declare a new object in the destination memory space and initialize it
+appropriately. For overload resolution, the parameter type must be an exact
+match in order for overload resolution to succeed since no conversions will be
+valid.
 
 Allowed:
-```c++
+```hlsl
+void doesSomething(inout uint4);
+void doesNothing(uint4);
+
 void fn(groupshared uint4 A) {
   float4 LocalA = (float4) A;
   doesSomething(LocalA);
+  doesSomething(A);
+  doesNothing(LocalA);
+  doesNothing(A);
   A = (uint4) LocalA;
 }
 ```
@@ -66,10 +75,6 @@ void fn2() {
 }
 ```
 
-DXC and Clang have implemented this feature in their 202x language modes, and
-allowed it to be used as an extension (with diagnostic issued) in earlier
-language modes.
-
 The feature supports any type that is valid to store in groupshared memory to be
 used as groupshared arguments, including user-defined data types.
 
@@ -81,8 +86,10 @@ is an obvious alternative. This proposal introduces a slightly conflicting
 syntax from what we would prefer with reference types available.
 
 This more minimal feature has material benefit today for both DXC and Clang, and
-can avoid Clang requiring special case handling for library functions. As such,
-this proposal is preferred to waiting until references can be finalized.
+can avoid Clang requiring special case handling for library functions that
+operate on `groupshared` memory such as the `Interlocked` intrinsics for atomic
+operations. As such, this proposal is preferred to waiting until references can
+be finalized.
 
 ### Overloading behavior
 
--- a/proposals/NNNN-groupshared-arguments.md
+++ b/proposals/NNNN-groupshared-arguments.md
@@ -13,8 +13,7 @@ params:
 
 |   | DXC     | Clang    |
 |---|---------|----------|
-| `groupshared` arguments (DXIL) | Complete | Complete |
-| `groupshared` arguments (SPIRV) | Complete | Complete |
+| `groupshared` arguments | Complete | Complete |
 
 ## Introduction
 
@@ -36,11 +35,10 @@ blocks of groupshared memory.
 ## Proposed solution
 
 HLSL 202x will allow the `groupshared` type annotation keyword on function
-parameter declarations. The keyword when applied to a parameter declaration of
-type `T`, alters the qualified type of the parameter to a `groupshared T &`
+parameter declarations. When applied to a parameter declaration of
+type `T`, the keyword alters the qualified type of the parameter to a `groupshared T &`
 (a reference to `groupshared` memory of type `T`).
 
-```c++
 void fn(groupshared uint4 A) {}
 ```
 
@@ -60,12 +58,11 @@ void fn(groupshared uint4 A) {
 ```
 
 Not Allowed:
-```c++
 void fn(groupshared uint4 A) {}
 void fn2() {
   float4 B = 1.0.xxxx;
-  fn(B); // Error:
-  fn((uint4)B); // Error:
+  fn(B); // error: no matching function for call to 'fn'
+  fn((uint4)B); // error: no matching function for call to 'fn'
 }
 ```
 
@@ -89,7 +86,6 @@ this proposal is preferred to waiting until references can be finalized.
 
 ### Overloading behavior
 
-```c++
 void fn(groupshared uint shared);
 void fn(inout uint u);
 
@@ -103,7 +99,6 @@ void caller() {
 }
 ```
 
-```c++
 void fn(groupshared uint shared);
 void fn(uint u);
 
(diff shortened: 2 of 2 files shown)


proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+    - spall: Sarah Spall
+  status: Under Consideration
+---
+
+## Implementation Status
+
+|   | DXC     | Clang    |
+|---|---------|----------|
+| `groupshared` arguments (DXIL) | Complete | Complete |
+| `groupshared` arguments (SPIRV) | Complete | Complete |
```
pow2clk: We've never divided implementation by DXIL vs SPIRV output before. Given that it's all the same anyway, I see no reason to start now.
```suggestion
| `groupshared` arguments | Complete | Complete |
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+This capability is further used in the [DirectX Linear
+Algebra](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)
+extension, which adds functions on a class method that rely on passing in large
+blocks of groupshared memory.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared T &`
```
pow2clk: As written, this requires a comma after "The keyword". I think reorganizing the sentence would be simpler.

```suggestion
parameter declarations. When applied to a parameter declaration of
type `T`, the keyword alters the qualified type of the parameter to a `groupshared T &`
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+Not Allowed:
+```c++
+void fn(groupshared uint4 A) {}
+void fn2() {
+  float4 B = 1.0.xxxx;
+  fn(B); // Error:
+  fn((uint4)B); // Error:
```
pow2clk:
```suggestion
  fn(B); // error: no matching function for call to 'fn'
  fn((uint4)B); // error: no matching function for call to 'fn'
```

proposals/0014-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+memory space and initialize it appropriately. For overload resolution, the
+parameter type must be an exact match in order for overload resolution to
+succeed since no conversions will be valid.
+
+Allowed:
+```c++
+void fn(groupshared uint4 A) {
+  float4 LocalA = (float4) A;
+  doesSomething(LocalA);
+  A = (uint4) LocalA;
```
pow2clk, line 65: The above statement, "No implicit or explicit conversion can change the memory space of an object" might be misconstrued to disallow this. It might be worth adding an explanation as to why it isn't.

proposals/0014-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+perform such a conversion, a user must declare a new object in the destination
+memory space and initialize it appropriately. For overload resolution, the
+parameter type must be an exact match in order for overload resolution to
+succeed since no conversions will be valid.
+
+Allowed:
+```c++
+void fn(groupshared uint4 A) {
+  float4 LocalA = (float4) A;
+  doesSomething(LocalA);
```
pow2clk, line 61: It might be useful to declare this function so it's clear that what is expressly allowed is the ability to to resolve a call to a non-groupshared parameter function.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+DXC and Clang have implemented this feature in their 202x language modes, and
+allowed it to be used as an extension (with diagnostic issued) in earlier
+language modes.
```
pow2clk: This doesn't seem to belong in the proposed solution. Perhaps in the motivation section?

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Alternatives considered
+
+[Reference
+types](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0006-reference-types.md)
+is an obvious alternative. This proposal introduces a slightly conflicting
+syntax from what we would prefer with reference types available.
+
+This more minimal feature has material benefit today for both DXC and Clang, and
+can avoid Clang requiring special case handling for library functions. As such,
```
pow2clk: Do "library functions" refer to implementations like linalg.h or the existing "Interlocked*" intrinsics? I don't think the latter have been called by that term anywhere else and the linalg.h functions aren't really special case handling.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+  fn(Shared); // ambiguous
+
+  uint Local;
+  fn(Local); // Not ambiguous
+  fn(5); // Not ambiguous
+}
+```
+
+The above overload sets will result in an error that the call is ambiguous
+when the call site argument is a `groupshared` variable.  They will not result
```
pow2clk: That is the behavior, but if the value passed to the function is groupshared, it doesn't "exactly match" the default memory space of the second overload without the groupshared argument.

proposals/0014-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+A function annotated with either `export` or `[noinline]` will not be allowed to
+have function parameter declarations annotated with `groupshared`.  Doing so
+will produce an error.
+
+The argument to a `groupshared` function parameter must be a `groupshared`
+variable.  If it is not, an error will be produced.
+
+The argument to a `groupshared` function parameter must be of the same exact
+type as the function parameter.  No implicit or explicit conversions are
+allowed.  If they are not exactly the same, an error will be produced.
```
pow2clk, line 124: Are these limitations? They seem to be explicit features of the spec that are described above. I see a limitation as something that might be changed later. Do we intend to allow implicit conversions to resolve overload matches at some later time?

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared T &`
+(a reference to `groupshared` memory of type `T`).
+
+```c++
+void fn(groupshared uint4 A) {}
+```
+
+No implicit or explicit conversion can change the memory space of an object. To
```
pow2clk: Given that no memory space conversions should be allowed, it would seem like converting from groupshared to the default memory space for overload resolution should be disallowed, yet this is: 

https://hlsl.godbolt.org/z/aeex3h4bK

I bring this up in order to suggest that we should consider if this is the desired behavior and we should specify it or that we should change the spec to match this behavior.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+blocks of groupshared memory.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared T &`
+(a reference to `groupshared` memory of type `T`).
+
+```c++
```
pow2clk: I know you dislike the hlsl parsing, but when using `groupshared` it actually does determine if its highlighted as a keyword or not.
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+void fn(groupshared uint4 A) {
+  float4 LocalA = (float4) A;
+  doesSomething(LocalA);
+  A = (uint4) LocalA;
+}
+```
+
+Not Allowed:
+```c++
```
pow2clk:
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+is an obvious alternative. This proposal introduces a slightly conflicting
+syntax from what we would prefer with reference types available.
+
+This more minimal feature has material benefit today for both DXC and Clang, and
+can avoid Clang requiring special case handling for library functions. As such,
+this proposal is preferred to waiting until references can be finalized.
+
+### Overloading behavior
+
+```c++
```
pow2clk:
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,137 @@
...
+
+void caller() {
+  fn(Shared); // ambiguous
+
+  uint Local;
+  fn(Local); // Not ambiguous
+}
+```
+
+```c++
```
pow2clk:
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
llvm-beanz: I split these initially because it was different work items to make it work, but since both are completed it's reasonable to collapse them. That said, I expect that to not be uncommon with DXC for features in general due to how little code the DXIL and SPIRV paths contain, and I think providing granular status is useful to our community so we should be okay with that int he future.

proposals/NNNN-groupshared-arguments.md
llvm-beanz: This isn't converting the address space, it is a lvalue-to-rvalue cast (effectively a load), which converts from a _reference to `groupshared` memory_ to a loaded local value. This actually creates a temporary of the object.

A concrete example of what this is talking about would be attempting to convert a reference to `groupshared` memory into a reference to device or thread local memory without creating a copy.

I can take a stab at clarifying this language.

proposals/NNNN-groupshared-arguments.md
llvm-beanz: This is talking about the "Interlocked" functions. I can clarify.

proposals/NNNN-groupshared-arguments.md
llvm-beanz: LValue-To-Rvalue conversions are ranked as "exact match". This is the current behavior in DXC, which is codified in the proposed language in the [overload resolution proposal](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0003-overload-resolution.md).

proposals/0014-groupshared-arguments.md
llvm-beanz: This callout is because this is different from how the `inout` and `out` keywords work in HLSL today.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,139 @@
...
+void caller() {
+  fn(Shared); // ambiguous
+
+  uint Local;
+  fn(Local); // Not ambiguous
+  fn(5); // Not ambiguous
+}
+```
+
+The above overload sets will result in an error that the call is ambiguous
```
bogner: Can we move the text here above the examples and word it in a way that explains the behaviour? Maybe:
> Overload sets including value types and `groupshared` arguments of the same underlying type are ambiguous if the call site argument is `groupshared`.

bogner (approved):
LGTM, minor editorial comments inline

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,127 @@
...
+blocks of groupshared memory.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. When applied to a parameter declaration of
+type `T`, the keyword alters the qualified type of the parameter to a `groupshared T &`
+(a reference to `groupshared` memory of type `T`).
+
+```
```
pow2clk: ?
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,127 @@
...
+  doesSomething(LocalA);
+  doesSomething(A);
+  doesNothing(LocalA);
+  doesNothing(A);
+  A = (uint4) LocalA;
+}
+```
+
+Not Allowed:
+```
```
pow2clk:
```suggestion
```hlsl
```

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,127 @@
...
+operate on `groupshared` memory such as the `Interlocked` intrinsics for atomic
+operations. As such, this proposal is preferred to waiting until references can
+be finalized.
+
+### Overloading behavior
+
+Overload sets including value types and `groupshared` arguments of the same
+underlying type are ambiguous if the call site argument is `groupshared`.
+
+```
```
pow2clk:
```suggestion
```hlsl
```

pow2clk (commented):
Don't know why the language type was removed from most of the code blocks.

llvm-beanz:

PR #74, opened 2026-06-19
[New] Add proposal for `const` member functions

llvm-beanz:
This adds a proposal for `const`-qualified non-static member functions.

### DIFF
--- a/proposals/0015-const-member-functions.md
+++ b/proposals/0015-const-member-functions.md
@@ -1,5 +1,5 @@
 ---
-title: NNNN - const-qualified Non-static Member Functions
+title: 0015 - const-qualified Non-static Member Functions
 params:
   authors:
   - llvm-beanz: Chris Bieneman
--- a/proposals/NNNN-const-member-functions.md
+++ b/proposals/NNNN-const-member-functions.md
@@ -8,6 +8,12 @@ params:
   status: Under Consideration
 ---
 
+## Implementation Status
+
+|   | DXC     | Clang    |
+|---|---------|----------|
+| `const` member functions | [Prototype implementation](https://github.com/llvm-beanz/DirectXShaderCompiler/commit/aa44718f81c58d6aca7d53d0705c897bd2c7eae4) | Complete |
+
 ## Introduction
 
 HLSL does not currently support `const` non-`static` member functions for user-defined data
@@ -63,6 +69,8 @@ call `()` operator:
 struct Pupper {
   void Wag() const { /* body omitted */ }
   void operator() const { /* body omitted */ }
+  void Walk() mutable { /* body omitted */ }
+  mutable int Speed;
 };
 ```
 
--- a/proposals/NNNN-const-member-functions.md
+++ b/proposals/NNNN-const-member-functions.md
@@ -1,0 +1,150 @@
+---
+title: NNNN - const-qualified Non-static Member Functions
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
+  status: Under Consideration
+---
+
+## Introduction
+
+HLSL does not currently support `const` non-`static` member functions for user-defined data
+types. This proposal seeks to add support for `const` non-`static` member functions, and to
+adopt const-correct behaviors across the HLSL library objects.
+
+## Motivation
+
+The absence of `const` non-`static` member functions causes some challenges since HLSL
+injected ASTs do have `const` and non-`const` non-`static` member functions.
+Further, since variables can be `const`-qualified, without the ability to
+specify `const` there are some cases that cannot be worked around without
+breaking `const`-correctness.
+
+DXC has received a number of issues relating to this. Two recent issues are
+listed below:
+
+* https://github.com/microsoft/DirectXShaderCompiler/issues/4340
+* https://github.com/microsoft/DirectXShaderCompiler/issues/4706
+
+In the first issue, a user defined data type's functions are basically unusable
+if the data type is placed in a `ConstantBuffer`. This results from the
+`ConstantBuffer` access returning `const &` objects.
+
+The second issue describes a similar problem, where overloaded operators on
+instances of `const`-qualified user-defined types are unusable.
+
+## Proposed solution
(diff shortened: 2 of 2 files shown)


proposals/0015-const-member-functions.md
```diff
@@ -0,0 +1,158 @@
...
+argument so it can only be applied to non-`static` member functions. See the
+examples below defining both a constant function and a constant overload of the
+call `()` operator:
+
+```c++
+struct Pupper {
+  void Wag() const { /* body omitted */ }
+  void operator() const { /* body omitted */ }
+  void Walk() mutable { /* body omitted */ }
+  mutable int Speed;
```
pow2clk, line 73: The addition of a usable `mutable` keyword is not mentioned in any of the normative language nor in any dependency. If it is the intent to introduce it alongside the rest of this, it needs to be described there. Otherwise, it should be removed from the example.

proposals/0015-const-member-functions.md
```diff
@@ -0,0 +1,158 @@
...
+they cannot remove the `const` qualifier and shall not convert from a constant
+lvalue to a non-constant rvalue by copying the implicit argument as is valid for
+other arguments.
+
+### HLSL Data Type Changes
+
+Introducing constant member functions provides an opportunity to revisit the
+const-correctness patterns of existing HLSL data types. With this change we will
+perform an audit of existing data types to provide constant and non-constant
+member functions as appropriate for the data type.
```
pow2clk, line 99: Am I correct that this spec will not be considered refined until this audit has been performed and the results explicitly included?

proposals/0015-const-member-functions.md
```diff
@@ -0,0 +1,158 @@
...
+  Hat H;
+};
+
+export int GetFeatherCount() {
+  return H.getFeathers();
+}
+```
+
+This code is valid under HLSL 2021 because HLSL ignores the const-ness of the
+implicit object parameter, however any writes to the constant data are dropped.
```
pow2clk, line 130: I'd be mildly curious to see how the dropping behavior can be observed. I could only get DXC to crash and clang to error.

proposals/0015-const-member-functions.md
```diff
@@ -0,0 +1,158 @@
...
+On introducing constant member functions, this code is ill-formed because these
+declarations inside a `cbuffer` are implicitly constant and will produce a
+diagnostic.
```
pow2clk, line 133: I haven't thought it through, but it might be worth considering making member functions of objects in a cbuffer default to const to minimize backward compatibility. Obviously if they try to modify their contents, then that will fail, but maybe disallowing such functions for cbuffer objects would be the least evil option.
```suggestion
On introducing constant member functions, this code is ill-formed because these
declarations inside a `cbuffer` are implicitly constant and will produce a
diagnostic when used to invoke non-constant member functions.
```

proposals/0015-const-member-functions.md
```diff
@@ -0,0 +1,158 @@
...
+
+Introducing constant member functions provides an opportunity to revisit the
+const-correctness patterns of existing HLSL data types. With this change we will
+perform an audit of existing data types to provide constant and non-constant
+member functions as appropriate for the data type.
+
+When applied to HLSL intangible types, the `const` qualifier will apply as if to
+the handle, not the data the handle grants access to. For example, a `const
+RWBuffer<T>` will still allow writes to the underlying resource, however the
+resource variable itself cannot be re-assigned.
```
pow2clk, line 104: That makes sense and at least rhymes with the behavior of const pointers. It seems though that const resources should not be assignable to non-const. Would you agree?

```hlsl
Texture2D tx1;

export void doit() {
    const Texture2D t = tx1;
    Texture2D t2 = t;
}
```

I ask because clang allows this: https://hlsl.godbolt.org/z/M5acY6xzE

proposals/0015-const-member-functions.md
llvm-beanz: There is no suggested normative language in this proposal yet, but I will take the feedback to clearly call out that the `mutable` keyword is an expected addition.

proposals/0015-const-member-functions.md
llvm-beanz: That is my current intent, but this is a draft initial proposal and it may change as the proposal is developed.

proposals/0015-const-member-functions.md
llvm-beanz: This is completely fine. The `const` applies to the name `t`. You can initialize a new name `t2` from a `const` value which allows `t2` to be reassigned, but `t` cannot be re-assigned.

Consider similar C++ code:
```c++
static int X = 0;

void fn() {
    const int Y = X;
    int Z = Y;
}
```

`Y` cannot be re-assigned because it is `const`, but its value may be used to initialize non-const variables.

proposals/0015-const-member-functions.md
llvm-beanz: I'm not sure how you would reasonably make a member function implicitly default to const. Doing that only on objects that happen to be used in cbuffers would likely cause some really odd user experiences. 

Consider for example a case where a structure is used both in and outside a cbuffer?
What if the use inside a cbuffer is actually unused in the shader so it gets eliminated in the optimizer?

This sounds like a path to some pretty gnarly implicit behaviors.

llvm-beanz: The suggestion here doesn't make sense. It seems like you're updating this to a more general statement, but the sentence is specifically talking about the code sample above becoming invalid, not just "when used to invoke a non-constant member".

bogner (approved):
This LGTM for the purposes of bringing it to under consideration. As the proposal itself notes there is a lot of the language that needs to be audited for the impact and implications of this before we could really accept it though.

proposals/0015-const-member-functions.md
pow2clk: Given that `const` on resources makes not the handle, but the contents unwritable, I consider this to be more analogous:

```c++
static int *X = 0;

void fn() {
    const int *Y = X;
    int *Z = Y;
}
```

llvm-beanz:

PR #119, opened 2026-06-24
[Lex.Ident] Add Identifiers subclause

### LINKED ISSUE #107: [Lex.Ident] Identifiers
Draft language for the Identifiers subclause.

llvm-beanz:
This adds the text for the identifiers subclause of the Lexical Conventions clause, and the required normative references to the Universal Character Sets (Unicode) standard and Unicode Standard Annexes for character classifications.

Resolves #107

### DIFF
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -247,8 +247,8 @@ in \gls{UAX44}. The characters which have the property are defined in
 \p An XID\_Start character is a character of the translation character set whose
 corresponding code point in \gls{UTF} has the XID\_Start property. An
 XID\_Continue character is a character of the translation character set whose
-corresponding code point in \gls{UTF} has the XID\_Continue. An identifier shall
-conform to Normalization Form C as specified in \gls{UTF}.
+corresponding code point in \gls{UTF} has the XID\_Continue property. An identifier
+shall conform to Normalization Form C as specified in \gls{UTF}.
 
 \p Some identifiers are reserved for use by implementations of this standard,
 but are not required to emit a diagnostic:
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -214,7 +214,52 @@ next token is a \textit{scalar-element-sequence} (\ref{Lex.Literal.Vector}). In
 this situation the \textit{pp-number} token is truncated to end before the
 period.
 
-%TODO: \Sec{Identifiers}{Lex.Ident}
+\Sec{Identifiers}{Lex.Ident}
+
+\begin{grammar}
+  \define{identifier}\br
+  identifier-start\br
+  identifier identifier-continue
+
+  \define{identifier-start}\br
+  nondigit\br
+  \textnormal{an element of the translation character set of class XID\_Start}
+
+  \define{identifier-continue}\br
+  digit\br
+  nondigit\br
+  \textnormal{an element of the translation character set of class XID\_Continue}
+
+  \define{nondigit} \textnormal{one of}\br
+  \terminal{a b c d e f g h i j k l m}\br
+  \terminal{n o p q r s t u v w z y z}\br
+  \terminal{A B C D E F G H I J K L M}\br
+  \terminal{N O P Q R S T U V W X Y Z \_}
+
+  \define{digit} \textnormal{one of}\br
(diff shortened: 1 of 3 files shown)


spec/lex.tex
```diff
@@ -214,7 +214,52 @@
...
+
+\p The XID\_Start and XID\_Continue properties are Derived core properties defined
+in \gls{UAX44}. The characters which have the property are defined in
+\gls{UAX31}.
+
+\p An XID\_Start character is a character of the translation character set whose
+corresponding code point in \gls{UTF} has the XID\_Start property. An
+XID\_Continue character is a character of the translation character set whose
+corresponding code point in \gls{UTF} has the XID\_Continue. An identifier shall
+conform to Normalization Form C as specified in \gls{UTF}.
```
Alexander-Johnston:
```suggestion
corresponding code point in \gls{UTF} has the XID\_Continue property. An identifier
shall conform to Normalization Form C as specified in \gls{UTF}.
```

spec/lex.tex
```diff
@@ -214,7 +214,52 @@
...
+
+\p Some identifiers are reserved for use by implementations of this standard,
+but are not required to emit a diagnostic:
+
+\begin{itemize}
+  \item Any identifier that contains a double underscore (\texttt{\_\_}), or
+  begins with an underscore (\texttt{\_}) followed by a capital letter is
+  reserved for any use by an implementation.
+  \item Any identifier that begins with an underscore (\texttt{\_}) is reserved
+  for any implementation to use as a name in the global namespace.
```
Alexander-Johnston, line 261: I just want to confirm my understanding/reading of these items with you.

In all contexts `__identifier` and `_Capital letter identifier` are reserved.
In exclusively the global namespace `_identifer` is reserved, but not in other contexts?
So things like
```
struct foo {
  int _a;
  ...

void bar() {
  int _a;
  ...
```
are still valid?

llvm-beanz: Yep. This matches C++'s current rules, which I think is safe for us to align on.

Alexander-Johnston: Perfect, thank you.

Alexander-Johnston (approved):
LGTM (and my non-exhaustive understanding of Unicode). Built locally and the pdf looks good.

jessebarker (approved):

jessebarker:
Out of curiosity, and because it came up in another standards meeting, we don't actually have to prohibit things like emojis, and other unicode representable characters, do we?  I'm assuming the explicit digit/non-digit definitions cover us, but thought I'd ask.

llvm-beanz:
> Out of curiosity, and because it came up in another standards meeting, we don't actually have to prohibit things like emojis, and other unicode representable characters, do we? I'm assuming the explicit digit/non-digit definitions cover us, but thought I'd ask.

This is a good question.

The way that the spec here is written is very similar with C and C++ which both rely on the Unicode annexes.

The official classifications come from [here](https://www.unicode.org/reports/tr31/#D1). A abbreviated quote (bolding mine):

>|Properties | General Description of Coverage|
> |-- | --|
>|ID_Start | ID_Start characters are derived from the Unicode General_Category of uppercase letters, lowercase letters, titlecase letters, modifier letters, other letters, letter numbers, plus Other_ID_Start, minus Pattern_Syntax and Pattern_White_Space code points...|
> | XID_Start | XID_Start characters are derived from ID_Start... |
> |ID_Continue | ID_Continue characters include ID_Start characters, plus characters having the Unicode General_Category of nonspacing marks, spacing combining marks, decimal number, connector punctuation, plus Other_ID_Continue, minus Pattern_Syntax and Pattern_White_Space code points...|
>|XID_Continue | XID_Continue characters are derived from ID_Continue... **XID_Continue characters are also known simply as Identifier Characters, because they are a superset of the XID_Start characters.**|

Basically unicode took the way many programming languages codify identifier character sets (digit/non-digit) and generalized that cross the full space of unicode characters so that programming languages can just point to unicode's specs and allow variables in any native language characters (or emoji).

farzonl (approved):

llvm-beanz:

PR #120, opened 2026-06-24
[Website] Add a SWAG progress bar

llvm-beanz:
This PR adds a script to generate an SVG image of a progress bar to help us estimate and visualize the progress we're making on the specification.

The current tracking is based on my completely wild guess that our first edition will land somewhere around 600 pages. This is probaly not an unreasonable guess, but it is a guess.

As we evolve our process and tracking we can iterate on this to make it more refined and with higher confidence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
 
 [![Documentation Build Status](https://github.com/hlsl-tc57/tc57/actions/workflows/hugo.yml/badge.svg)](https://github.com/hlsl-tc57/tc57/actions/workflows/hugo.yml)
 
+SWAG progress estimate:
+<img src="https://hlsl-tc57.github.io/tc57/spec/assets/progress_bar.svg">
+
 This repository documents and tracks the specification and evolution process for
 the HLSL programming language under Ecma Technical Committee 57 (TC57). This
 repository is made public to provide visibility into the feature development
--- a/scripts/progress_bar.py
+++ b/scripts/progress_bar.py
@@ -6,6 +6,7 @@ import re
 import subprocess
 import sys
 from datetime import date, timedelta
+from typing import Optional
 
 
 DEFAULT_START_DATE = date(2026, 1, 1)
@@ -43,8 +44,7 @@ def estimate_completion(progress: int, start_date: date, today: date) -> str:
     completion_date = start_date + timedelta(days=total_days)
     return completion_date.strftime("%B %-d, %Y")
 
-
-def generate_progress_bar_svg(progress: int, start_date: date = DEFAULT_START_DATE, today: date = None) -> str:
+def generate_progress_bar_svg(progress: int, start_date: date = DEFAULT_START_DATE, today: Optional[date] = None) -> str:
     """Generate an SVG progress bar for the given progress percentage."""
     if not 0 <= progress <= 100:
         raise ValueError(f"Progress must be between 0 and 100, got {progress}")
--- a/scripts/progress_bar.py
+++ b/scripts/progress_bar.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+"""Generate an SVG progress bar image."""
+
+import argparse
+import re
+import subprocess
+import sys
+from datetime import date, timedelta
+
+
+DEFAULT_START_DATE = date(2026, 1, 1)
+
+
+def count_pdf_pages(path: str) -> int:
+    """Return the total page count of a PDF using pdfinfo."""
+    try:
+        result = subprocess.run(
+            ["pdfinfo", path],
+            capture_output=True, text=True, check=True,
+        )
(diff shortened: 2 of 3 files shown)


llvm-beanz:
The current state of this progress bar will look something like this:
tmp

README.md
```diff
@@ -4,6 +4,9 @@
 
 [![Documentation Build Status](https://github.com/hlsl-tc57/tc57/actions/workflows/hugo.yml/badge.svg)](https://github.com/hlsl-tc57/tc57/actions/workflows/hugo.yml)
 
+SWAG progress estimate:
+<img src="https://hlsl-tc57.github.io/tc57/spec/assets/progress_bar.svg">
```
farzonl, line 8: can we get an image of this in the PR description.  Also what are your thoughts on taking snapshots of this picture and committing them so we can see progress over time.

scripts/progress_bar.py
```diff
@@ -0,0 +1,212 @@
...
+    days_elapsed = (today - start_date).days
+    if days_elapsed <= 0:
+        return "N/A (start date is today or in the future)"
+    total_days = days_elapsed * 100 / progress
+    completion_date = start_date + timedelta(days=total_days)
+    return completion_date.strftime("%B %-d, %Y")
+
+
+def generate_progress_bar_svg(progress: int, start_date: date = DEFAULT_START_DATE, today: date = None) -> str:
+    if not 0 <= progress <= 100:
```
farzonl, line 48: you did python style doc strings for `estimate_completion` and `count_pdf_pages` why not this function?

README.md
llvm-beanz: I put the image in a comment here:
https://github.com/hlsl-tc57/tc57/pull/120#issuecomment-4791357794

In terms of snapshotting, I think it might be useful to combine that with the discussion of monthly updates. We could have each monthly update generate the image and record that in the updates which will each be archived.

scripts/progress_bar.py
```diff
@@ -0,0 +1,212 @@
...
+    bar_width = 500
+    bar_height = 30
+    total_width = bar_x + bar_width + 8
+    label_y = bar_y + bar_height + 16
+    total_height = label_y + 6
+
+    # Tick mark settings: ticks rise upward from the bottom of the bar
+    tick_y_bottom = bar_y + bar_height
+
+    filled_width = bar_width * progress / 100
```
farzonl, line 67: should this be divided by 100.0? you do `.2f`  in all the `lines.append` that use filled_width and i'm wondering if that means this shouldn't be integer division?

scripts/progress_bar.py
```diff
@@ -0,0 +1,212 @@
...
+        return today.strftime("%B %-d, %Y") + " (complete)"
+    days_elapsed = (today - start_date).days
+    if days_elapsed <= 0:
+        return "N/A (start date is today or in the future)"
+    total_days = days_elapsed * 100 / progress
+    completion_date = start_date + timedelta(days=total_days)
+    return completion_date.strftime("%B %-d, %Y")
+
+
+def generate_progress_bar_svg(progress: int, start_date: date = DEFAULT_START_DATE, today: date = None) -> str:
```
farzonl: I don't think the default should be None. it seems like this would violate type checking. Strict type checkers would flag today: date = None as an error. It  would be better to use the union operator  or Optional from the typing module.

scripts/progress_bar.py
llvm-beanz: I think the completion date is such a wild guess that if we have some rounding error it almost certainly won't impact it much.

scripts/progress_bar.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+DEFAULT_START_DATE = date(2026, 1, 1)
+
+
+def count_pdf_pages(path: str) -> int:
+    """Return the total page count of a PDF using pdfinfo."""
+    try:
+        result = subprocess.run(
+            ["pdfinfo", path],
```
farzonl: I dunno that this is a better solution than to use `PyPDF2` or `PyMuPDF`. It means we need `pdfinfo` binary which is fine if this just works in the workflow but it also means that folks on our team on windows can't test the progress_bar.py script on their windows dev boxes unless you want to support a `pdfinfo.exe` string and add some if windows os checks.

farzonl (approved):

damyanp (approved):

scripts/progress_bar.py
llvm-beanz: We need the pdf tools to generate the pdfs from latex anyways, so this doesn't really introduce a dependency.

llvm-beanz: Oh! Nvm. I thought pdftools came as a dependency of TexLive, but it doesn't seem to.

PR #122, opened 2026-06-24
[New] Specify ternary operator behavior in HLSL

### LINKED ISSUE #123: [Expr.Cond] HLSL must specify assignment operands that are ternary expressions
In DXC, with the 2021 language version, C semantic validation logic is used when handling ternary operators as operands to assignment operators.
C semantic validation rejects such constructs with a diagnostic error.
However, in the 202x language version, DXC uses C++ semantic validation logic, accepting this construct.
This inconsistency in behavior is not specified.
HLSL should specify exactly how it handles ternary operators when used as lvalues, within assignment operators.
This is necessary so that the Clang compiler implements the right behavior going forward.

bob80905:
This PR specifies ternary operator behavior in HLSL, specifically when the result of the operator is an operand of the assignment operator.

Fixes https://github.com/hlsl-tc57/tc57/issues/123

### DIFF
--- a/proposals/0017-hlsl-ternary-operator-behavior.md
+++ b/proposals/0017-hlsl-ternary-operator-behavior.md
@@ -1,5 +1,5 @@
 ---
-title: NNNN - HLSL Ternary Operator Behavior
+title: 0017 - HLSL Ternary Operator Behavior
 params:
     authors:
     - bob80905: Joshua Batista
--- a/proposals/NNNN-hlsl-ternary-operator-behavior.md
+++ b/proposals/NNNN-hlsl-ternary-operator-behavior.md
@@ -41,7 +41,7 @@ operands are `lvalue` `DeclRefExpr` nodes, while the result of the ternary
 itself is an rvalue (note the absence of an `lvalue` annotation on the
 `ConditionalOperator`). This was produced with DXC in HLSL 202x mode
 (`-T lib_6_5 -HV 202x -ast-dump`,
-[Compiler Explorer](https://godbolt.org/z/r1sv16ToM)):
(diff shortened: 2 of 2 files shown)


proposals/0017-hlsl-ternary-operator-behavior.md
```diff
@@ -0,0 +1,95 @@
...
+validating ternary operators, while HLSL 202x currently falls back into the
+C++ semantic analysis path. As a result, the same source compiles differently in
+DXC depending on the selected language version.
+
+Consider the following example, compiled in DXC with both HLSL 2021 and HLSL
+202x ([Compiler Explorer](https://hlsl.godbolt.org/z/j3TcE8Y38)):
+
+```hlsl
+export int fn(bool B, int X, int Y) {
+    (B ? X : Y) += 1;
```
devshgraphicsprogramming, line 65: fun fact, this is illegal in SPIR-V and really hard to legalize, current legalizations of such things inflate compile time by 10x on DXC as found by @AnastaZIuk because loops need unrolling and functions inlining.

KHR_variable_pointers will NOT save you here.

The real solution is to "fix" SPIR-V IMHO not cater to its weird limitations.

llvm-beanz: This is fundamentally equivalent to:
```
if (B)
  X += 1;
else
  Y += 1;
```

Which is definitely not illegal in SPIRV. If DXC is doing something dumb here that's a bug.

bogner (approved):

llvm-beanz (commented):
Approved to merge as _Under Consideration_ in the 2026-07-21 meeting.

PR #133, opened 2026-06-29
[Lex.Keywords] Add Keywords subclause

llvm-beanz:
This subcluase lists all the reserved keywords. This will need revising as decisions are made for other language proposals which impact the list of keywords, but for now this is representative of the keywords that are respected by DXC.

### DIFF
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -216,7 +216,28 @@ period.
 
 %TODO: \Sec{Identifiers}{Lex.Ident}
 
-%TODO: \Sec{Keywords}{Lex.Keywords}
+\Sec{Keywords}{Lex.Keywords}
+
+\begin{grammar}
+  \define{keyword} \textnormal{any of}\br
+  \terminal{auto bool break case cbuffer center centroid class column\_major}\br
+  \terminal{const constexpr continue discard do double else enum export}\br
+  \terminal{false float for globallycoherent groupshared if in indices}\br
+  \terminal{inline inout int interface line lineadj linear namespace nointerpolation}\br
+  \terminal{noperspective operator out packoffset payload point precise}\br
+  \terminal{primitives reordercoherent return row\_major sampler\_state shared sizeof}\br
+  \terminal{snorm static struct switch tbuffer template this triangle}\br
+  \terminal{triangleadj true typedef typename uniform unorm unsigned using vertices}\br
+  \terminal{void while}
+\end{grammar}
+
+% The above list includes the interface, shared, and uniform keywords which are
+% proposed to be removed in 0004.
+
+\p The identifiers defined in the grammar above are reserved as keywords.
+During the phases of translation when preprocessing tokens are converted to
+tokens, keywords are unconditionally treated as keywords except when they appear
+in attribute specifiers (\ref{Decl.Attributes}).
 
 %TODO: \Sec{Operators and Punctuators}{Lex.Operators}
 
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -231,8 +231,8 @@ period.
   \terminal{void while}
 \end{grammar}
 
-% The above list includes the interface, shared, and uniform keywords which are
-% proposed to be removed in 0004.
+% The above list includes the interface, sampler_state, shared, and uniform
+% keywords which are proposed to be removed in 0004.
 
 \p The identifiers defined in the grammar above are reserved as keywords.
 During the phases of translation when preprocessing tokens are converted to


llvm-beanz:
A question for the TC here:

Do we want to respect DXC's behavior of treating some of these keywords as context-dependent?

Specifically, DXC does not treat the following as keywords in all contexts:

* precise
* sample
* globallycoherent
* reordercoherent
* center
* indices
* vertices
* primitives
* payload

My gut feeling is that `globallycoherent`, `reordercoherent` and `precise` should be globally reserved keywords, and the remaining ones should be context-dependent (meaning we only treat them as keywords in specific locations in the grammar, and allow users to define them as identifiers elsewhere).

jessebarker:
> A question for the TC here:
> 
> Do we want to respect DXC's behavior of treating some of these keywords as context-dependent?
> 

I don't think I love the idea of the language saying "this *might* be a keyword, but it might not".
I also can't find references in any of the current documentation to most of those examples.  Are they actually documented somewhere currently?

llvm-beanz:
> I don't think I love the idea of the language saying "this _might_ be a keyword, but it might not". I also can't find references in any of the current documentation to most of those examples.

I tend to agree, but I worry a bit about `sample`, `center`, `indices`, `vertices`, `primitives` and `payload`, which all seem like words that might show up in shaders as identifiers. `sample` itself is a mess because we use it both as a keyword and as an identifier within behavior that well want to standardize 😢.

> Are they actually documented somewhere currently?

I can't find any documentation on `sample` or `center`, which are keywords that can be applied to input fields to change the interpolation behavior.

`indices`, `vertices`, `primitives` and `payload` are part of mesh shader support which is documented [here](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html).

spec/lex.tex
```diff
@@ -216,7 +216,28 @@
...
+  \terminal{inline inout int interface line lineadj linear namespace nointerpolation}\br
+  \terminal{noperspective operator out packoffset payload point precise}\br
+  \terminal{primitives reordercoherent return row\_major sampler\_state shared sizeof}\br
+  \terminal{snorm static struct switch tbuffer template this triangle}\br
+  \terminal{triangleadj true typedef typename uniform unorm unsigned using vertices}\br
+  \terminal{void while}
+\end{grammar}
+
+% The above list includes the interface, shared, and uniform keywords which are
+% proposed to be removed in 0004.
```
farzonl: FYI for anyone looking for porposal 4: https://github.com/hlsl-tc57/tc57/pull/134

farzonl (approved):

spec/lex.tex
```diff
@@ -216,7 +216,28 @@
...
+  \terminal{auto bool break case cbuffer center centroid class column\_major}\br
+  \terminal{const constexpr continue discard do double else enum export}\br
+  \terminal{false float for globallycoherent groupshared if in indices}\br
+  \terminal{inline inout int interface line lineadj linear namespace nointerpolation}\br
+  \terminal{noperspective operator out packoffset payload point precise}\br
+  \terminal{primitives reordercoherent return row\_major sampler\_state shared sizeof}\br
+  \terminal{snorm static struct switch tbuffer template this triangle}\br
+  \terminal{triangleadj true typedef typename uniform unorm unsigned using vertices}\br
+  \terminal{void while}
+\end{grammar}
```
pow2clk, line 232: For reference information only: We discussed a bit offline the strange cases here where DXC does not treat some of these keywords as such.  They can be treated as identifiers. I've determined that `center`, `constexpr`, `globallycoherent`, `indices`, `payload`, `precise`, `primitives`, `reordercoherent`, and `vertices` can all be used as variable names. Even in contexts where they can be used as type qualifiers such as parameters will allow other parameters to use the keywords as identifiers.

The chances of these being used as variable names in existing shaders varies. It would sure be simpler to just treat these as keywords going forward. It could break some shaders, but they would all produce errors. Maintaining the peculiar behavior of having them serve both as qualifiers and variable names seems surprising and potentially confusing to users as well and tricky to maintain. I think we should make them all keywords with standard behavior.

https://godbolt.org/z/hvsK4c6fP

spec/lex.tex
pow2clk: `sampler_state` will be removed as part of the effects syntax removal as well. We should make a note to flesh out the syntax removal section of proposal 0004.
```suggestion
% The above list includes the interface, sampler_state, shared, and uniform
% keywords which are proposed to be removed in 0004.
```

pow2clk: https://github.com/hlsl-tc57/tc57/blob/main/proposals/0004-202x-removals.md

pow2clk (approved):
Approved though I recommend accepting my small addition to the comment.

spec/lex.tex
llvm-beanz: I had an agenda item on the 2026-07-07 meeting that we didn't get to and has been carried over to the 2026-07-21 meeting to discuss how we want to approach writing the detailed design for the removals in proposal 0004.

Since features that don't exist aren't actually going to have detailed designs (we're not designing anything) we'll need to ask the committee for guidance how to proceed.

llvm-beanz:

PR #135, opened 2026-06-30
[Lex.Operators] Add lexing rules for operators

llvm-beanz:
The notable difference between HLSL and C/C++ here is that I've excluded digraph and trigraph tokens. I'm sure we don't need those right?

### DIFF
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -224,10 +224,12 @@ period.
   \define{preprocessing-op-or-punc}\br
     preprocessing-operator\br
     operator-or-punctuator
+    
   \define{preprocessing-operator}\br
     \terminal{\#}\br
     \terminal{\#\#}
-  \define{operator-or-punctuator}\textnormal{one of}\br
+    
+  \define{operator-or-punctuator} \textnormal{one of}\br
     \terminal{\{ \} [ ] ( ) ; : ...}\br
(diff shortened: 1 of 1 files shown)


spec/lex.tex
```diff
@@ -218,7 +218,31 @@
...
+\Sec{Operators and Punctuators}{Lex.Operators}
+
+\begin{grammar}
+  \define{preprocessing-op-or-punc}\br
+    preprocessing-operator\br
+    operator-or-punctuator
+  \define{preprocessing-operator}\br
+    \terminal{\#}\br
+    \terminal{\#\#}
+  \define{operator-or-punctuator}\textnormal{one of}\br
```
Alexander-Johnston:
```suggestion
    operator-or-punctuator
    
  \define{preprocessing-operator}\br
    \terminal{\#}\br
    \terminal{\#\#}
    
  \define{operator-or-punctuator} \textnormal{one of}\br
```
Fixes the `\define` being pushed onto the prior line and keeps the formatting in line with similar sections.

Alexander-Johnston (changes requested):

Alexander-Johnston (approved):

pow2clk (approved):

llvm-beanz:

PR #143, opened 2026-07-21
[Basic.Lookup] Specify name lookup

### LINKED ISSUE #95: [Basic.Lookup] Name Lookup
We need specification language for name lookup and the differentiation between qualified and unqualified names and the lookup rules for each.

llvm-beanz:
This PR fleshes out the name lookup subclause of our specification including covering unqualified name lookup, argument-dependent name lookup, qualified name lookup, elaborated type specifiers, and using directives.

Resolves #95

### DIFF
--- a/spec/basic.tex
+++ b/spec/basic.tex
@@ -154,7 +154,132 @@ either;
 \p Scopes nest within each other. A \textit{containing scope}, is a scope that
 fully contains another scope which is said to be \textit{contained}.
 
+\p The \textit{outermost scope} of all declarations is the global scope, while
+the \textit{innermost scope} of a declaration is the scope the declaration
+appears within.
+
 \Sec{Name Lookup}{Basic.Lookup}
+
+\p Name lookup unambiguously identifies one or more declarations associated with
+a name, the name may only refer to more than one declaration if the name is a
+function name in which case the set of declarations form an overload set
+(\ref{Overload.Decl}). When applicable, overload resolution takes place after
+name lookup.
+
+\p Name lookup in the context of an expression is unqualified name lookup in the
+enclosing scope of the expression.
+
+\Sub{Unqualified Name Lookup}{Basic.Lookup.Unqualified}
+
+\p For unqualified name lookup, scopes are searched in a specified order from
+the innermost scope to the outermost scope, and lookup ends as soon as a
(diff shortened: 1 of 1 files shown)


spec/basic.tex
```diff
@@ -154,8 +154,317 @@
...
+and namespaces for each function argument will be searched. The set of
+associated classes and namespaces for a given argument is constructed by the
+following:
+
+\begin{itemize}
+\item If the type of the argument is a fundamental type the set of classes and
+the set of namespaces are empty.
+\item If the argument is of class type, the set of associated classes will
+include the class of the argument, it's direct and indirect base classes, and
+any classes it is a member of. The set of associated namespaces will include the
```
Alexander-Johnston, line 239: Do you have an example of what should be included by `any classes it is a member of` here?
On first reading this seemed to me as if
```
class Foo ...

class Bar {
  Foo foo;
  ...
```
Would lead to Bar being added to the set of class type args, when I think this is meant to refer to
```
class Foo {
  class Bar {
  ...
```
I see that this is C++ ADL rules/wording, but I can't see anywhere that we've defined what a class being a member of a class is. (Perhaps worth coming back here later to add links to other sections as they're written?)

I suppose the first case is more specifically 'A member variable of a class' while the second is 'A member of a class', but I do wonder if the C++ wording can be improved upon here.

Unless you think I am looking too far ahead and tripping up on nothing?

llvm-beanz: You have the right interpretation in that class `A` is a member of class `B` if class `A`'s declaration is inside class `B`. Such as:

```
class B {
  class A {...};
};
```

This means that A's fully qualified name is B::A (because `A` is a member of `B`.

By contrast your initial interpretation:

```
class A {...};
class B {
  A varOfA;
};
```

In this case `A` is not a member of `B`, but `varOfA` is. This is because the point of declaration of `A` is outside `B`, but the point of declaration of `varOfA` is inside `B`.

Alexander-Johnston (approved):
LGTM, thanks for the clarification.

The examples are very helpful. I wouldn't say no to adding more in the future, but I'm happy for this to go in as it is now

llvm-beanz:
Approved in the 2026-08-04 meeting.

PR #151, opened 2026-07-27
[New] Add proposal for hlsl::size_t

llvm-beanz:
This proposal seeks to add a size type to HLSL to allow the standard libarary to be defined against an implementation-defined size type so that offsets and sizes may vary by implementation.


jessebarker (approved):

Alexander-Johnston (approved):

llvm-beanz:
Approved to merge as **Under Consideration** in the 2026-08-18 meeting.

PR #165, opened 2026-08-04
[Basic.Scope] Elaborate on types of scopes

llvm-beanz:
This fleshes out the remainder of the Basic.Scope cluase.

### DIFF
--- a/spec/basic.tex
+++ b/spec/basic.tex
@@ -153,7 +153,7 @@ either;
 
 \p Scopes nest within each other. A \textit{containing scope}, is a scope that
 fully contains another scope which is said to be \textit{contained}. The
-smallest containing scope of a \texttt{S} is the \textit{parent scope} of
+smallest containing scope of \texttt{S} is the \textit{parent scope} of
 \texttt{S}, except for template parameter scopes, where the parent scope will be
 the smallest containing scope that is not a template parameter scope.
 
@@ -189,20 +189,20 @@ after the \textit{identifier}, and the name is declared to be a
 \textit{class-name} in the scope containing the forward declaration. Otherwise,
 if the \textit{elaborated-type-specifier} is used in a
 \textit{decl-specifier-seq} or \textit{parameter-declaration-clause} during a
-function declaration, the the name is declared to be a \textit{class-name} in
+function declaration, then the name is declared to be a \textit{class-name} in
 the namespace that contains the function declaration.
 
 \p The point of declaration for an \textit{injected-class-name} is immediately
 following the opening brace of the definition of the class definition.
 
-\p The point of declaration of a template parameter is immediately after it's
+\p The point of declaration of a template parameter is immediately after its
 \textit{template-parameter}.
 
 \Sub{Block Scope}{Basic.Scope.Block}
 
 \p A block scope is introduced by either a \textit{compound-statement}, or a
 selection or iteration statement (\ref{Stmt.If}, \ref{Stmt.Switch},
-\ref{Stmt.For}, \ref{Stmt.While}), or a substatement such a statement.
+\ref{Stmt.For}, \ref{Stmt.While}), or a substatement of such a statement.
 
 \p A name declared inside a block is said to have \textit{block-scope} and is
 local to that block. Its potential scope begins at its point of definition and
@@ -237,10 +237,10 @@ have namespace scope.
 
 \p The potential scope of a namespace member name begins at its point of
 declaration and contains the remainder of the namespace scope through the end of
-the translation unit. Each \textit{using-declaration} nominating a namespace
+the translation unit. Each \textit{using-directive} nominating a namespace
 extends the potential scope of namespace members in the nominated namespace to
 include the portion of the potential scope containing the
-\textit{using-declaration} following the member's point of declaration.
+\textit{using-directive} following the member's point of declaration.
 
 \p A member of a namespace may also be referred to as a qualified name any point
 after its point of declaration via the rules defined in
@@ -258,7 +258,7 @@ name's point of declaration and extends to the end of the translation unit.
 \Sub{Class Scope}{Basic.Scope.Class}
 
 \p The declarative region of a class is defined by the curly braces (\texttt{\{
-\}}) enclosing it's optional \textit{member-specialization}. Declarations within
+\}}) enclosing its optional \textit{member-specialization}. Declarations within
 a class's \textit{member-specialization} are said to be \textit{members} of that
 class. Any names introduced by these declarations are said to be \textit{member
 names} of the class, and they have class scope.
@@ -274,7 +274,7 @@ its use as when re-evaluated with the completed class.
 \Sub{Enumeration Scope}{Basic.Scope.Enumeration}
 
 \p The name of a scoped enumerator (\ref{Decl.Enum}), has enumeration scope.
-It's potential scope begins at its point of declaration and terminates at the
+Its potential scope begins at its point of declaration and terminates at the
 end of the enumerator's declaration.
 
 \Sub{Template Parameter Scope}{Basic.Scope.Template}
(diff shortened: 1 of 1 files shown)


spec/basic.tex
```diff
@@ -152,7 +152,143 @@
 \end{itemize}
 
 \p Scopes nest within each other. A \textit{containing scope}, is a scope that
-fully contains another scope which is said to be \textit{contained}.
+fully contains another scope which is said to be \textit{contained}. The
+smallest containing scope of a \texttt{S} is the \textit{parent scope} of
```
jessebarker: nit: I think you want to leave out the "a" here?

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+\p The name of a class member may be looked up in a class immediately after its
+point of declaration, even if the containing class is incomplete.
+
+\p The point of declaration for a name introduced in an
+\textit{elaborated-type-specifier} that is a forward declaration is immediately
+after the \textit{identifier}, and the name is declared to be a
+\textit{class-name} in the scope containing the forward declaration. Otherwise,
+if the \textit{elaborated-type-specifier} is used in a
+\textit{decl-specifier-seq} or \textit{parameter-declaration-clause} during a
+function declaration, the the name is declared to be a \textit{class-name} in
```
jessebarker: nit: "..., then the name is...."

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+following the opening brace of the definition of the class definition.
+
+\p The point of declaration of a template parameter is immediately after it's
+\textit{template-parameter}.
+
+\Sub{Block Scope}{Basic.Scope.Block}
+
+\p A block scope is introduced by either a \textit{compound-statement}, or a
+selection or iteration statement (\ref{Stmt.If}, \ref{Stmt.Switch},
+\ref{Stmt.For}, \ref{Stmt.While}), or a substatement such a statement.
```
jessebarker: nit: "...substatement of such a..."

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+
+\p A name declared inside a block is said to have \textit{block-scope} and is
+local to that block. Its potential scope begins at its point of definition and
+ends at the end of the block. A variable declared within a block is called a
+\textit{local variable}.
+
+\p The potential scope of a function parameter name begins at its point of
+declaration. A parameter name shall not be redeclared in the outermost block of
+the function definition.
```
jessebarker, line 214: This is saying that the parameter name is local to the function definition "block", correct?

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+said to have \textit{global namespace scope}, and are said to be \textit{global
+names}.
+
+\p The potential scope of a name with global namespace scope begins at the
+name's point of declaration and extends to the end of the translation unit.
+
+\Sub{Class Scope}{Basic.Scope.Class}
+
+\p The declarative region of a class is defined by the curly braces (\texttt{\{
+\}}) enclosing it's optional \textit{member-specialization}. Declarations within
```
jessebarker: nit: "its" (possessive)

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+bodies, default arguments, and initializers of non-static data members in the
+class.
+
+\p A name used in a class shall refer to the same declaration at the point of
+its use as when re-evaluated with the completed class.
+
+\Sub{Enumeration Scope}{Basic.Scope.Enumeration}
+
+\p The name of a scoped enumerator (\ref{Decl.Enum}), has enumeration scope.
+It's potential scope begins at its point of declaration and terminates at the
```
jessebarker: nit: "Its" (possessive)

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+end of the enumerator's declaration.
```
KStocky: Do we mean that it ends at the end of the **enumeration's** declaration? From this I am taking that:

enumerator = named constant
enumeration = type that defines a list of enumerators


With this definition, this reads to me like the following should not be valid:
```hlsl
enum class A
{
    B,
    C = B
};
```
Because after the declaration of B (an enumerator), its scope ends, so we should need to qualify access to it with `A::B` even while inside the enumeration scope.

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+
+\Sub{Template Parameter Scope}{Basic.Scope.Template}
+
+\p Each \textit{template-parameter} and \textit{template-declaration} introduces
+a template parameter scope. A \textit{template-parameter}'s introduced scope
+contains the smallest \textit{template-parameter-list} containing the
+\textit{template-parameter}. A \textit{template-declaration}'s introduced scope
```
KStocky: This wording makes me wonder if the following would be valid:

```hlsl
template<typename T>
struct Wrapper
{
    T Object;
};
```

It kinda reads like the template parameter `T` introduces its own scope and it is contained within that scope. This to me would imply that `T` is not visible outside of the template parameter list.

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
...
+declarative regions for each \textit{namespace-definition} with that name.
+
+\p Declarations within a \textit{namespace-body} declare entities that are said
+to be \textit{members} of the namespace. Any names introduced by these
+declarations are said to be \textit{member names} of the namespace, and they
+have namespace scope.
+
+\p The potential scope of a namespace member name begins at its point of
+declaration and contains the remainder of the namespace scope through the end of
+the translation unit. Each \textit{using-declaration} nominating a namespace
```
KStocky: This sounds like we are conflating using-declarations and using-directives.

```hlsl
using namespace abc; // using-directive
using abc::myVar; //using-declaration
```

I think we should be saying `using-directive` here rather than `using-declaration`

spec/basic.tex
```diff
@@ -152,7 +152,143 @@
 \end{itemize}
 
 \p Scopes nest within each other. A \textit{containing scope}, is a scope that
-fully contains another scope which is said to be \textit{contained}.
+fully contains another scope which is said to be \textit{contained}. The
+smallest containing scope of a \texttt{S} is the \textit{parent scope} of
+\texttt{S}, except for template parameter scopes, where the parent scope will be
+the smallest containing scope that is not a template parameter scope.
+
+\p A declaration's \textit{target scope} is the scope its point of declaration
```
KStocky, line 160: I believe that all definitions are also declarations. So in the following example:
```hlsl
struct A
{
    void Do();
};

void A::Do()
{
}
```

The declaration of the out-of-class definition of `A::Do()` is lexically located in the global namespace, but its target scope is `A`.

Do we need to explicitly account for qualified declarations here?

llvm-beanz: Out-of-class definitions are actually weird because they're not allowed to introduce names (you can't declare a member outside a class without declaring it inside the class first).Those rules are captured later in the C++ spec in the Classes clause (https://timsong-cpp.github.io/cppwp/n4868/class.mfct).

I don't believe the C++ specification has any additional wording around scoping since the behavior falls out from needing to be initially declared inside the class's declaration.

KStocky: Ok fair enough!

spec/basic.tex
llvm-beanz: Yea, there's an oddity in the way scopes are defined for function parameters and declarations in loop or selections statements. Basically there is a scope for the function prototype itself, then a secondary scope for the outermost block of the function, there's a special case rule here that names declared in the function prototype can't be re-declared in the outermost block. A concrete example is:

```
void fn(int X) {
  int X; // Can't do this!
}

void fn(int X) {
  if (X > 5) {
    int X; // This is fine 🔥
  }
}
```

spec/basic.tex
llvm-beanz: I've adjusted the wording to use the grammar construct explicitly, hopefully that reads clearly.

KStocky: Yup, that reads better to me

spec/basic.tex
llvm-beanz: Template parameter scoping is so weird!!! I added some clarification that the template parameter scope extends to the full template declaration's scope.

KStocky: Yes I think this is fine now. And I would agree. It took me a while to see that the wording of this was to accommodate `template template` parameters... Or at least I assume that that is the intent.

jessebarker (approved):

llvm-beanz:
Approved by consensus to merge this PR at the 2026-008-18 meeting.

PR #168, opened 2026-08-11
[0010] Narrow down proposed features

llvm-beanz:
This narrows down to three sets of features from C++ from 8 papers from C++ 11, 14, and 17. These features are all low-hanging fruit that are easy to implement in DXC, already working in Clang, and provide meaningful value to our users and compiler implementers.

The features are:
* Templates
  - Variadic templates
  - Simplified parsing of closing angle brackets
  - Template variable declarations
* Type deduction
  - `auto` and `decltype` keywords
  - Type-deduced return types for normal functions (C++14 `auto`)
* Compile-time expressions
  - `constexpr`
  - `static_assert` (C++17)

### DIFF
--- a/proposals/0010-modern-cxx.md
+++ b/proposals/0010-modern-cxx.md
@@ -8,10 +8,9 @@ params:
 
 * Issues:
    * auto: [#24](https://github.com/microsoft/hlsl-specs/issues/24)
-   * decltype: (#82)[https://github.com/microsoft/hlsl-specs/issues/82)
+   * decltype: [#82](https://github.com/microsoft/hlsl-specs/issues/82)
    * constexpr: [#21](https://github.com/microsoft/hlsl-specs/issues/21)
-(#74)[https://github.com/microsoft/hlsl-specs/issues/74]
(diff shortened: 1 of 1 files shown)


proposals/0010-modern-cxx.md
```diff
@@ -47,6 +46,71 @@ their career after modern C++ was widely adopted.
...
+
+### Compile-time evaluation
+
+C++11 introduced a set of features which subsequent versions have built on to
+enable compile-time evaluation of portions of programs and compile-time
+verification.
+
+HLSL will adopt the `constexpr` keyword and the associated behaviors introduced
+to C++11 in the paper [Generalized constant expressions
+(n2235)](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2235.pdf).
```
bogner, line 97: Might be worth reiterating that this doesn't include a lot of the expanded power of `constexpr` in C++14 and later. This does specifically point out that this is the C++11 feature here, but that may be a subtle point to readers.

bogner (approved):

PR #175, opened 2026-08-25
[0013] Named casts - detailed design

### LINKED ISSUE #136: [0013] Expose `static_cast` in `hlsl` namespace
**Which proposal does this relate to?**
[0013 - named casts](https://github.com/hlsl-tc57/tc57/blob/main/proposals/0013-named-casts.md)

**Describe the issue or outstanding question.**
@bogner suggested exposing `static_cast` both in and outside the `hlsl` namespace.

**Additional context**
https://github.com/hlsl-tc57/tc57/pull/72#pullrequestreview-4639007024

llvm-beanz:
Flesh out the detailed design for named casts. This includes rolling back #136 to address the more complicated nature of `static_cast`.

Resolves #136.


proposals/0013-named-casts.md
```diff
@@ -48,3 +48,94 @@ in the `hlsl` namespace aligning with modern C++ conventions to avoid additions
...
+```
+
+_Requires:_ `sizeof(DestTy) == sizeof(SrcTy)` is `true`;
+_Returns:_ An object of type `DestTy`. Each bit in the the object representation
+of the `Src` argument is replicated into the cooresponding bit in the value
+representation of the newly created `DestTy` object. Padding bits in the result
+are unspecified. If an object or subobject of the returned value has a bit
+pattern that does not map to a value of that type, the behavior is undefined.
+
+  \item \textit{Effects:} the actions performed by the function.
```
Alexander-Johnston: These need filled out still, no?

proposals/0013-named-casts.md
```diff
@@ -48,3 +48,94 @@ in the `hlsl` namespace aligning with modern C++ conventions to avoid additions
...
+In the detailed design below the HLSL `static_cast` cannot produce lvalue
+outputs, this is because lvalue references in HLSL are not spellable. This
+causes some cascading reductions and simplifications of the spec language and
+reduces the scope of `static_cast`.
+
+## Detailed Design
+
+### Addition to [Lex.Keywords]
+
+The keyword `static_cast` is added to the grammar of keywords.
```
jessebarker: I don't think I feel strongly, but this reads a little oddly to me.  Maybe "The identifier `static_cast`..." or "The token `static_cast`..."?  Maybe I'm being too pedantic in trying to say "it's not a keyword at the beginning of the sentence, but it is at the end." :-D

proposals/0013-named-casts.md
llvm-beanz: This is a just me moving too fast. I copied these from the LaTeX to have as a template to fill out as I documented each API, then I left them behind :(. Will remove.

proposals/0013-named-casts.md
llvm-beanz: I'm not sure what the best way to word this is, but I agree as it is sounds confusing. The intent is really that we're just adding `static_cast` to the table here:

https://hlsl-tc57.github.io/tc57/spec/5-Lex.html#Lex.Keywords

I'll try to reword this to make that more clear.

proposals/0013-named-casts.md
```diff
@@ -24,13 +24,13 @@ values from a source type to a destination type with tight validation.
...
 
-HLSL will introduce three new named cast templates, `hlsl::static_cast<T>`,
+HLSL will introduce three new named cast templates, `static_cast<T>`,
 `hlsl::elementwise_cast<T>` and `hlsl::bit_cast<T>`.
 
 The `hlsl::static_cast<T>(V)` will behave similar to the C++ `static_cast<T>(V)`
 expression, except as changes are required due to other differences between C++
 and HLSL. It will convert the expression `V` to the result type `T`. The main
-difference from C++ is that HLSL's `hlsl::static_cast<T>` will always produce an
+difference from C++ is that HLSL's `tatic_cast<T>` will always produce an
```
llvm-beanz:
```suggestion
difference from C++ is that HLSL's `static_cast<T>` will always produce an
```

proposals/0013-named-casts.md
```diff
@@ -48,3 +48,89 @@ in the `hlsl` namespace aligning with modern C++ conventions to avoid additions
...
+#### Function Template `bit_cast`
+
+```
+template<typename DestTy, typename SrcTy>
+constexpr DestTy bit_cast(const SrcTy Src);
+```
+
+_Requires:_ `sizeof(DestTy) == sizeof(SrcTy)` is `true`;
+_Returns:_ An object of type `DestTy`. Each bit in the the object representation
+of the `Src` argument is replicated into the cooresponding bit in the value
```
bogner, line 133:
```suggestion
of the `Src` argument is replicated into the corresponding bit in the value
```

proposals/0013-named-casts.md
```diff
@@ -48,3 +48,89 @@ in the `hlsl` namespace aligning with modern C++ conventions to avoid additions
...
+causes some cascading reductions and simplifications of the spec language and
+reduces the scope of `static_cast`.
+
+## Detailed Design
+
+### Addition to [Lex.Keywords]
+
+> The keyword `static_cast` is added to the grammar of keywords. Updated
+> rendering below, with addition highlighted.
+![Latex Rendering](0013-assets/KeywordsUpdates.png)
```
bogner, line 65: I'd suggest simply:
> The `static_cast` keyword is added to `[Lex.Keywords]`

I also think that if we want anything supplementary to that then just showing the change itself might be clearer than the image. It's a bit janky because of whitespace, but I think this is pretty legible:

```diff
--- a/spec/lex.tex
+++ b/spec/lex.tex
@@ -264,14 +264,14 @@
 \Sec{Keywords}{Lex.Keywords}

 \begin{grammar}
   \define{keyword} \textnormal{any of}\br
   \terminal{auto bool break case cbuffer center centroid class column\_major}\br
   \terminal{const constexpr continue discard do double else enum export}\br
   \terminal{false float for globallycoherent groupshared if in indices}\br
   \terminal{inline inout int line lineadj linear namespace nointerpolation}\br
   \terminal{noperspective operator out packoffset payload point precise}\br
   \terminal{primitives reordercoherent return row\_major sizeof}\br
-  \terminal{snorm static struct switch tbuffer template this triangle}\br
-  \terminal{triangleadj true typedef typename unorm unsigned using vertices}\br
-  \terminal{void while}
+  \terminal{snorm static static_cast struct switch tbuffer template this}\br
+  \terminal{triangle triangleadj true typedef typename unorm unsigned using}\br
+  \terminal{vertices void while}
 \end{grammar}
```

bogner (approved):
Typo fix and some minor editorial opinions on how to say we've added a keyword. LGTM.

PR #176, opened 2026-08-25
[0004] Add 202x removals language to spec

llvm-beanz:
This integrates the language from the 202x removals proposal into the specification. There are a few editorial formatting modifications:

* URLs are printed in full text in parenthesis rather than as titled links so that printed documents will preserve them.
* `cbuffer` declarations was changed to _cbuffer-declarations_ referencing a grammar formation.
* `interface` declarations was changed to _interface declarations_ referencing a concept that will not appear in the final spec.
* Updated D3D 10 links to D3D 11.
* Added `technique` and `technique11` keywords.


spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+
+\p These features are not included in the standard, nor are they conforming
+extensions, as such conforming implementation will not support these features.
+
+\Sub{Effects For Direct3D}{AnnexB.LegacyDX.Effects}
+
+\p The Effects runtime support in Direct3D 9 through Direct3D 11 relied on a set of
+syntax extensions for HLSL. The latest documentation for the Direct3D 11 Effects
+APIs and shader syntax is available on Microsoft's public documentation
+site (\url{https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-reference-effect}).
```
pow2clk: https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d11-graphics-programming-guide-effects is the more recent and has more information on HLSL rather than the D3D API. I don't feel strongly about it, but you may want to use that link instead of this.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+
+\p The main additions to HLSL for the Effects feature are the annotation
+syntax (\url{https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-effect-annotation-syntax})
```
pow2clk: should choose to use the d3d11 documentation above, the corresponding section could be used here: https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d11-effect-annotation-syntax

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+	pass RenderPass
+	{
+		SetVertexShader(CompileShader(vs_2_0, VSMain()));
+		SetPixelShader(CompileShader(ps_2_0, PSMain()));
+	}
+}
+\end{HLSL}
+
+\p The effects syntax has three keywords: \texttt{sampler\_state},
+\texttt{technique10}, and \texttt{pass}.
```
pow2clk: technique<version> is the set of keywords https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d11-effect-technique-syntax DXC accepts `technique`, `technique10`, and `technique11` as ignored keywords. I think this should be rewritten to encompass all of them and not just this one.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
 resource declaration.
+
+\Sec{Legacy DirectX Features}{AnnexB.LegacyDX}
+
+\p The existing reference compilers support a set of features that worked in
+tandem with the Direct3D runtime or other support libraries which are no longer
+supported in the latest versions of Direct3D.
+
+\p These features are not included in the standard, nor are they conforming
+extensions, as such conforming implementation will not support these features.
```
pow2clk, line 96: The change is flagged as normative and the wording here certainly sounds normative in that an implementation that does support these would be considered nonconforming. Is that the intent? I didn't think normative text belonged in annex B. I'd like an answer to the questions either way, but I don't really like that exclusions are a requirement for conformant implementations. A hypothetical implementation might include some extensions that make use of some of this syntax. I don't think that should be disallowed.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+the Direct3D runtime could swap interface implementations at runtime enabling
+code reuse and specialization.
+
+\p This feature is not supported in Direct3D 12, Vulkan, nor any actively
+supported shader bytecode. DXC's implementation of the \texttt{interface}
+keyword is limited to static verification that objects inheriting from the
+interface provide definitions for all interface members.
+
+\p A conforming implementation of this standard will not support the
+\textit{interface} keyword.
```
pow2clk, line 144: This repeats what is said at the top, but it isn't repeated in the effects subsection. Is this one unnecessary or does each subsection need this language? I think one or the other should be done for consistency.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+
+\Sec{Legacy HLSL Features}{AnnexB.LegacyHLSL}
+
+\p The actively supported reference compiler DXC has support for parsing some
+HLSL constructs that had meaning and impact in FXC's implementation, but do
+nothing in DXC. These legacy language features are omitted from this standard.
```
pow2clk, line 150: I like the "omitted from the standard" language better than "conforming implementations will not". If this could be used instead above, I'd prefer it.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+\textit{cbuffer-declarations} into shader reflection information. This
+information was used by the Effects for Direct3D
+(\url{https://github.com/Microsoft/FX11}) runtime, but could also be used
+directly by users.
+
+\p The DXC compiler supported parsing initializers but did not capture them into
+the shader reflection or any other metadata. This standard disallows
+initializing variables in \textit{cbuffer-declarations}, an initializer in such
+a declaration is ill-formed and a conforming implementation must issue a
+diagnostic.
```
pow2clk, line 164: This however seems to go beyond "omitted from the standard". Are all these requirements meant to mean the same thing? This one requires a diagnostic.

spec/annex-b.tex
```diff
@@ -85,3 +85,87 @@
...
+
+\Sub{\texttt{uniform} and \texttt{shared} keyword}{AnnexB.LegacyHLSL.Keywords}
+
+\p The current implementation of DXC parses but ignores the \texttt{uniform} and
+\texttt{shared} keywords which had meaning in earlier versions of HLSL. A
```
pow2clk, line 169: `shared` is an effects keyword. I think it belongs in that section. https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax

pow2clk (commented):
My main concern is the strict prohibition of certain syntax in a conforming implementation. If we were to list everything that should be invalid syntax, the document would be very long. Granted these are a special case because they were allowed before, but this is not what I understood the intent of Annex B to be. I think we should state what is required more than what is forbidden.

spec/annex-b.tex
llvm-beanz: The PR is flagged as "normative" because it contains changes to the normative language in the "Lexical Conventions" clause.

The text in Annex B is non-normative. Saying that a conforming implementation will not support these features doesn't need to be treated as normative here because the "Conformance" clause already makes this requirement (although worded differently).

spec/annex-b.tex
llvm-beanz:
```suggestion
site (\url{https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d11-graphics-programming-guide-effects}).
```

spec/annex-b.tex
llvm-beanz:
```suggestion
syntax (\url{https://learn.microsoft.com/en-us/windows/win32/direct3d11/d3d11-effect-annotation-syntax})
```

spec/annex-b.tex
llvm-beanz:
```suggestion
\p The effects syntax has five keywords: \texttt{sampler\_state},
\texttt{technique}, \texttt{technique10}, \texttt{technique11}, and
\texttt{pass}.
```

spec/annex-b.tex
llvm-beanz: This language was approved by the committee for integration to the spec at the last meeting. I don't believe it should be changed except for editorial corrections.

If we need to follow up on this, we should do it in a separate issue/PR.

llvm-beanz:
@pow2clk, I've integrated some of the changes you requested. This language was approved in the 2026-08-18 meeting for integration into the specification, I'm hesitant to make deeper changes in this PR. If we want to follow up in an issue or subsequent PR to address any further defects please file issues.

pow2clk (approved):

PR #177, opened 2026-08-25
[0016] Non-member operator overloading - detailed design

llvm-beanz:
This just pulls the detailed language for operator overloading out into the non-member operator overloading proposal from the overload resolution proposal.

The language here covers the intent of this proposal:

> An overloaded operator function must be either a non-static member
> function or a non-member function, and must have at least one
> parameter whose type is a class or enumeration.


proposals/0016-non-member-operator-overloading.md
```diff
@@ -74,6 +74,170 @@ source compatibility if they choose, but at their own maintenance expense. It
...
+
+```
+operator-function-id:
+  `operator` operator
+operator: one of
+  `+` `-` `*` `/` `%` `^` `&` `|` `~` `!` `<` `>` `<<` `>>` `<=` `>=` `&&` `||` `,` `()` `[]`
+```
+
+The operators `+`, `-`, `*`, and `&` may only be overloaded in their binary
+form.
```
jessebarker, line 89: No increment/decrement operators?

jessebarker (approved):
The language relocation seems fine.  Just a quick question about a couple of operators.

proposals/0016-non-member-operator-overloading.md
llvm-beanz: At the moment DXC doesn't support those, so I'm matching the current behavior.

I think the big question about extending this is that we don't have references, so the pre-increment can't actually return a reference to `this`.

PR #179, opened 2026-08-25
[Math.Conversions] Add Math Conversion Functions

llvm-beanz:
This PR adds the math conversion functions subclause of the Math functions library clause.

### DIFF
--- a/spec/math-conversions.tex
+++ b/spec/math-conversions.tex
@@ -1,0 +1,171 @@
+\Sec{Conversion Functions}{Math.ConversionFunctions}
+
+\p This subclause lists and describes conversion functions for converting values
+stored in numeric types and vectors of numeric types to other numeric types or
+vectors of such types.
+
+\begin{HLSL}
+template<uint N> vector<double, N> asdouble(vector<uint, N> lowBits, vector<uint, N> highBits);
+
+template<uint N> vector<float, N> asfloat(vector<float, N> input);
+template<uint N> vector<float, N> asfloat(vector<int, N> input);
+template<uint N> vector<float, N> asfloat(vector<uint, N> input);
+
+template<uint N> vector<float16\_t, N> asfloat16(vector<float16\_t, N> input);
+template<uint N> vector<float16\_t, N> asfloat16(vector<int16\_t, N> input);
+template<uint N> vector<float16\_t, N> asfloat16(vector<uint16\_t, N> input);
+
+template<uint N> vector<int16\_t, N> asint16(vector<float16\_t, N> input);
+template<uint N> vector<int16\_t, N> asint16(vector<int16\_t, N> input);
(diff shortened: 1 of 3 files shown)


spec/math-conversions.tex
```diff
@@ -0,0 +1,171 @@
...
+\end{itemize}
+
+\begin{HLSL}
+template<uint N> vector<uint16\_t, N> asuint16(vector<float16\_t, N> input);
+template<uint N> vector<uint16\_t, N> asuint16(vector<int16\_t, N> input);
+template<uint N> vector<uint16\_t, N> asuint16(vector<uint16\_t, N> input);
+\end{HLSL}
+
+\begin{itemize}
+  \item \textit{Returns:} A scalar or vector object of 32-bit unsigned integer
```
Alexander-Johnston:
```suggestion
  \item \textit{Returns:} A scalar or vector object of 16-bit unsigned integer
```

spec/math-conversions.tex
```diff
@@ -0,0 +1,171 @@
+\Sec{Conversion Functions}{Math.ConversionFunctions}
+
+\p This subclause lists and describes conversion functions for converting values
+stored in numeric types and vectors of numeric types to other numeric types or
+vectors of such types.
+
+\begin{HLSL}
+template<uint N> vector<double, N> asdouble(vector<uint, N> lowBits, vector<uint, N> highBits);
```
jessebarker, line 9: Are all of the `as<base_type>` cast functions defined somewhere else in the spec (totally possible I've missed it)?  I know they're in the general HLSL reference docs, but seems like something we should have at least a glossary definition for if we're using them to define these operations.

jessebarker: I see the operations described below, so that's handled.  I just wonder about using this shorthand before it is defined.

bogner (approved):
The text looks good to me, and I think any source compatibility concerns will point out programming mistakes so it is indeed best to treat them as errors.

Note though that this is dependent on a converation about `vector<T, 1>` - depending how we fall on whether `vector<T, 1>` and a scalar are interchangeable we might also need explicit scalar overloads. This is a broader issue though so I'll file an issue to discuss that separately.

jessebarker (approved):
LGTM

spec/math.tex
```diff
@@ -0,0 +1,6 @@
+\Ch{Mathmatics Functions}{Math}
```
llvm-beanz, line 1:
```suggestion
\Ch{Mathematics Functions}{Math}
```

Alexander-Johnston (approved):
LGTM

gmitrano-activision (approved):

llvm-beanz:
Approved to merge by consensus in the 2026-09-15 meeting.

PR #182, opened 2026-09-01
[0005] Integrate approved spec language

llvm-beanz:
This specification language was approved at the 2026-09-01 meeting for inclusion in the draft specification.


pow2clk (approved):
Looks fine. I created #194 to cover the other ways of declaring cbuffers at some point.

PR #192, opened 2026-09-08
[Website] Add slugs for permalinks

### LINKED ISSUE #191: [Website] Proposal permalink
It would be handy if proposals had a permalink created for their proposal number.

llvm-beanz:
This enbles us having proposals with permalinks so that even if we move the proposals around in the future the base URL + `proposal/NNNN` will work

Fixes #191


alsepkow (approved):

website/hugo.toml
```diff
@@ -20,6 +20,9 @@ ignoreLogs = ['warning-goldmark-raw-html']
     endLevel = 3
     ordered = false
 
+[permalinks]
```
alsepkow, line 23: This will change the published URLs. Do we care about adding aliases to make sure existing published URLs still work?

llvm-beanz: I don't know that we've published the URLs anywhere that matters yet, but part of why I want to setup permalinks is so that we can rely on them being stable. I fully expect that at some point in the future we'll want to archive completed proposals off to a different sub-directory in the git repo, and this gives us a way to make sure that URLs are stable.

We should also consider getting a domain at some point.

llvm-beanz: I can also confirm that it does effectively kill the old URLs.

